PR #18, opened 2019-08-31
Refactor monopackage into separate packages

patrickhulce:


paulirish:
the `server` component seemed very distinct from everything else.

are you thinking we can break that out as well?

patrickhulce:
> are you thinking we can break that out as well?

Already in a commit that wasn't pushed yet until I had wifi :) it was just significantly more complicated than splitting in the first place, so split the commits

paulirish:
hot

PR #35, opened 2019-10-28
feat: add commitedAt and ancestorCommittedAt

paulirish:


packages/cli/src/upload/upload.js
```diff
@@ -72,7 +72,7 @@ function getCurrentHash() {
   if (envVars.TRAVIS_PULL_REQUEST_SHA) return envVars.TRAVIS_PULL_REQUEST_SHA;
   if (envVars.TRAVIS_COMMIT) return envVars.TRAVIS_COMMIT;
 
-  const result = childProcess.spawnSync('git', ['rev-list', '--no-merges', '-n', '1', 'HEAD'], {
```
patrickhulce: does this impact things or just readability preference?

connorjclark: fwiw i find the new change to be more readable. one arg per array entry is good

packages/cli/src/upload/upload.js
```diff
@@ -82,6 +82,25 @@ function getCurrentHash() {
   return result.stdout.trim();
 }
 
+/**
+ * @param {string} currentHash
+ * @return {string}
+ */
+function getCommitTime(currentHash) {
+  const result = childProcess.spawnSync(
```
connorjclark: `execFileSync` is way better for this, as it just returns the stdout (as a buffer).

https://nodejs.org/api/child_process.html#child_process_child_process_execfilesync_file_args_options

paulirish:
> would you like some pointers or shall I takeover the PR?

please do take it over :)

packages/cli/src/upload/upload.js
patrickhulce: wait I'm a bit confused, why is `execFileSync` way better @connorjclark ? is it just that `result` is a string instead of accessing `result.stdout` to get the string?

I'm also confused why the try/catch is better. I'm not sure we want our users getting some obscure error directly from git and if we're remapping all of the errors, throwing on `result.status !== 0` seems like the more straightforward way to go

patrickhulce:
@googlebot I consent

packages/cli/src/upload/upload.js
connorjclark: tldr ok `spawnSync` is better here, since we want stdout and to check the error code w/o using `try/catch`.

I guess I just don't like `spawn*` :p I suppose the difference is trivial when you using `sync`. `spawn` is good for running a program that will output a ton of data (it will get streamed). `exec(File)` waits until all the data is done to do the `cb`, which is good when the program will output just one little line of text. Which again, doesn't matter for sync.

> I'm also confused why the try/catch is better.

wasn't meaning to suggest we use caught error message. my hate for `spawn` is stronger than my hate for `try/catch` i guess :)

patrickhulce: ok cool yeah just wanted to make sure I wasn't missing something :)

agreed that execFile is *much* easier for this scenario than spawn if it were all async 👍

PR #42, opened 2019-11-07
fix failure on merge-base

paulirish:
repro

```sh
mkdir blah
cd blah
git init
git remote add origin git@github.com:exterkamp/lighthouse-ci-action.git
git -c http.extraheader="AUTHORIZATION: basic ***" fetch --tags --prune --progress --no-recurse-submodules origin "+refs/heads/*:refs/remotes/origin/*"
git checkout --progress --force f2892d1121a20ba2ec047057e5351fdff7fa4905
```

```sh
git merge-base HEAD master
# fatal: Not a valid object name master
```


patrickhulce:
it feels like we should have a chain of fallbacks for some of these functions because depending on how the repo was cloned nothing is going to work 100% of the time

PR #54, opened 2019-11-10
more dots in the overview graph

paulirish:
this screenshot is more than twenty but i think its quite a lovely overview.

image

### DIFF
--- a/packages/server/src/ui/routes/project-dashboard/project-graphs.jsx
+++ b/packages/server/src/ui/routes/project-dashboard/project-graphs.jsx
@@ -179,7 +179,7 @@ export const ProjectGraphs = props => {
         .filter(build => build.branch === branch)
         .sort((a, b) => new Date(b.runAt).getTime() - new Date(a.runAt).getTime())
         .map(build => build.id)
-        .slice(0, 8),
+        .slice(0, 20),
     [builds, branch]
   );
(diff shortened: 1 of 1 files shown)


packages/server/src/ui/routes/project-dashboard/project-graphs.jsx
```diff
@@ -179,7 +179,7 @@ export const ProjectGraphs = props => {
         .filter(build => build.branch === branch)
         .sort((a, b) => new Date(b.runAt).getTime() - new Date(a.runAt).getTime())
         .map(build => build.id)
-        .slice(0, 8),
+        .slice(0, 20),
```
patrickhulce: don't you need to change the `limit` value on the API call too? I wouldn't expect this alone to generate the graph there...

paulirish: hmm that's quite possible. i was doing localoverrides on the minified JS. it's very possible i also tweaked a `limit`. :)

patrickhulce: oh I fetch the most recent 20 already so this is fine but any *more* would require a limit change :)

the only issue here is this issues 12 more network requests the way the API is structured right now, so loading the graphs will be quite a bit slower with this. something to address later I suppose.

PR #55, opened 2019-11-10
git short hash of 7

paulirish:
7 characters is the typical short length (used in github and travis)

admittedly, git uses 9 these days with --short but I think its nicer to use the short one, and to align with github & travis.

### DIFF
--- a/packages/server/src/ui/routes/build-view/build-hash-selector.jsx
+++ b/packages/server/src/ui/routes/build-view/build-hash-selector.jsx
@@ -134,7 +134,7 @@ const BuildLineItem = props => {
           withDevLine={props.withDevLine}
         />
         <Pill variant={variant} avatar={build}>
-          <span className="build-hash-selector__hash">{build.hash.slice(0, 8)}</span>
+          <span className="build-hash-selector__hash">{build.hash.slice(0, 7)}</span>
         </Pill>{' '}
         <span className="build-hash-selector__commit">{build.commitMessage}</span>
         <span className="build-hash-selector__links">


connorjclark:
The length of shas shown by git varies with the size of the repo. It increases the length of the sha printed to guarantee it is unique - but only against commits made so far. No guarantees for future commits. see [this](https://blog.cuviper.com/2013/11/10/how-short-can-git-abbreviate/)

Linus suggested 12 for Linux, so that's probably OK to use. But just using 7, always, will likely break for big projects.

I ran that cool script in that link above for `lighthouse`:
```
68536 objects
 4: 44525 / 18446
 5: 4373 / 2160
 6: 300 / 150
 7: 20 / 10
 8: 0 / 0
0ae904082dce852660f6dd9a80be6c52ba7b80bf
0ae9040ef82a860fb3285c3b360923aa321d10c6
387c063643f7ea93efb3fcbef6c7571ecc3fc959
387c063a5aa44df4eb52d5e2706f8508ce9b9d94
48280ff17358208c3f51ded7ea51a0bcd3e87179
48280ff8ee6ccfaade8cfb843808a8c21243a5d7
56a29cf2c544c775c8fc7500cae9866d6a7b18bf
56a29cf9b4277c6ccbbc94f52f6837b65fc16dcb
6d0cd5b5f2c4adf1368c0af33257cded829ef7af
6d0cd5b788399ed41d5854427eb4df56741400cc
ad96042608420ce73caef69e4c55cb38ca99821f
ad96042d233bf0ae775964f8fbe7e1ab614a6a86
b3e593d29c868f46f3fb79da771d309782a240d7
b3e593d8a705bc6570ab51c4dd5ea3e1afc6ba59
c9480cb1da5a82903e369dd478cb500fbe6e7852
c9480cb649ad304e5366b1887b2bbd088722157b
deb3b35cd54b5cc34341f26438220c14261a5668
deb3b35df2f35c6a1dfbacaa54f913d511782ef3
e46e914351c68f79ad67897b53f093b77f9bd64a
e46e914ed3374f3fc0c1636ed9989e67d47304e7
```

This means that for Lighthouse, there are 20 commits that are ambiguous if you only look at the first 7 characters.

patrickhulce:
I have no strong preference here since this is entirely for display and collisions in this particular context are not very impactful, +1 to connor's statement that git dynamically decides the length

paulirish:
> This means that for Lighthouse, there are 20 commits that are ambiguous if you only look at the first 7 characters. And there are 10 unique shas of 7 characters.

we looked at this and turns out most of these hashes are OBJECTS not commits. :)

anyway. 

this thing is only for display. we'd have to nerdfight harder if this was about a URL or something important. :p

connorjclark:
1. I shared collisions of objects which is a bit unfair :)
1. for the UI (commit chips) I think 7 is great
1. for the URL slug I would suggest 12.

paulirish:
> 1. I shared collisions of objects which is a bit unfair :)
> 2. for the UI (commit chips) I think 7 is great
> 3. for the URL slug I would suggest 12.


:shipit:  SHIP IT

connorjclark:
buttttt

```
git rev-list --all --abbrev=0 --abbrev-commit |   awk '{ a[length] += 1 } END { for (len in a) print len, a[len] }'
4 2426
5 4132
6 445
7 23
8 1

```

there is one collision at 8 for lighthouse :nerd_face:

PR #59, opened 2019-11-11
Update autorun document location

roderickhsiao:
Link to correct document location

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -40,7 +40,7 @@ script:
 
 That's it! With this in place, you'll have Lighthouse reports collected and and asserted.
 
-See the [autorun docs](./cli#autorun.md) for more including running your own webserver and uploading the reports for later viewing.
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -40,7 +40,7 @@ script:
 
 That's it! With this in place, you'll have Lighthouse reports collected and and asserted.
 
-See the [autorun docs](./cli#autorun) for more including running your own webserver and uploading the reports for later viewing.
+See the [autorun docs](./cli#autorun.md) for more including running your own webserver and uploading the reports for later viewing.
```
patrickhulce:
```suggestion
See the [autorun docs](./cli.md#autorun) for more including running your own webserver and uploading the reports for later viewing.
```

I think this is what we want? :)

roderickhsiao: ah, good catch :) thanks

PR #72, opened 2019-11-14
Add README for @lhci/cli package

calvinf:
Add readme for the core `npm` package.  Probably worth doing for the other two, too, but this is a start.


calvinf:
@googlebot I signed it!

Issue #74, opened 2019-11-14
Document how to setup free CI server


HugoGresse:
Hi,

I've setup a [base PR](https://github.com/HugoGresse/open-feedback/pull/217) with Lighthouse CI on GitHub Action.  [Result](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1573724026011-28121.report.html).

But I'm laking some features: 
1. The diff against master branch. 
I though it was built-in and I'll be able to compare by current PR to the master branch for eg. (from [this repo screenshots](https://user-images.githubusercontent.com/39191/68522269-7917b680-025e-11ea-8d96-2774c0a0b04c.png))
My bet is that I can have this diff only with the Lighthouse Server behind. Do you think we could have a free-budget style for Lighthouse Server on OSS projects? 
2. Having a budget setup. I've setup the lighthouserc.json as follow: 
```
{
    "ci": {
        "budgetsFile": "./lighthouse/budget.json",
        "collect": {
            "staticDistDir": "./build"
        },
        "upload": {
            "target": "temporary-public-storage"
        }
    }
}
```
but so far nothing change. 

I'm having difficulties following the documentation overall. There a lot of option but it lack of full example like the cli doc but for the lighthouserc.json.

patrickhulce:
Thanks for filing @HugoGresse!

> My bet is that I can have this diff only with the Lighthouse Server behind.

Correct the historical reports and diffing are available only on the CI server.

> Do you think we could have a free-budget style for Lighthouse Server on OSS projects?

It's on our roadmap, but a large scale service is understandably a much larger effort than a service supporting a single organization. 

> Having a budget setup. I've setup the lighthouserc.json as follow:

Very close!

```json
{
  "ci": {
    "assert": {
        "budgetsFile": "./lighthouse/budget.json"
    },
    "collect": ...
  }
}
```

> I'm having difficulties following the documentation overall. There a lot of option but it lack of full example like the cli doc but for the lighthouserc.json.

We'll work on setting up richer examples of complete config files 👍

patrickhulce:
Oh, if you're just looking for a free hosting setup, Heroku's free plan fits the use case quite well since they come with free persistent postgresql databases. That's actually what we used for our canary testing environment to ensure it would work for folks on free-quality hardware :)

The idea of a free container through GCP is an interesting one, but not sure we'll be able to pull that off internally. We'll keep you posted. 

[...]

PR #76, opened 2019-11-14
Docs: Add note to Temporary Storage Server for diffing

HugoGresse:
I'm not sure if I'm the right guy to add the second part notice though it's written [here](https://github.com/GoogleChrome/lighthouse-ci/issues/74#issuecomment-554023063) 😁

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -223,7 +223,7 @@ lhci upload --target=temporary-public-storage
 
 If you're using `autorun`, setting `ci.upload.target` to `temporary-public-storage` in your `lighthouserc.json` is all that's necessary. The upload step will happen automatically.
 
-> The historical reports and diffing are available only with the Lighthouse CI server. Lighthouse mainteners are working on a Lighthouse CI server hosted on their side with free plan. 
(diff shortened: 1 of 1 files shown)


patrickhulce:
Many thanks for the PR @HugoGresse!

Is it possible you found out about Lighthouse CI through Chrome Dev Summit and that's where it seemed like you would get historical diffing without a server setup? :)

I ask because I think the docs themselves wouldn't really lead one to conclude you get historical diffing without the CI server and the sentence added in this PR almost exactly duplicates the sentence two lines below it. 

We also don't really want to broadcast a free service until we have concrete plans for one. It is just on our radar/roadmap but nothing substantial to report or commit to yet.

WDYT?

calvinf:
Maybe add a note to the "Collecting Lighthouse Results" section to say:

> Diffing between runs isn't available from `temporary-public-storage`; you have to run a LHCI server and setup your build to upload files there in order to get it.  See [Lighthouse CI Server docs](https://github.com/GoogleChrome/lighthouse-ci/blob/master/docs/getting-started.md#the-lighthouse-ci-server) for details.

HugoGresse:
I'm following some Google Developer account on twitter and find it there.  I found about the diffing on the repo screenshot in the main readme :D 

Calvin proposition seems to be a better fit, I've updated the PR, hope your using squash & merge.

Issue #48, opened 2019-11-08
thoughts on configuration


connorjclark:
# existing work

typescript doesn't have named a CLI option for setting the config file, but their config is named `tsconfig.json`

___

eslint uses `--config`, although it does have `--no-eslintrc`. allowed config names are many:
![image](https://user-images.githubusercontent.com/4071474/68508575-94b19b80-0223-11ea-8f45-921dae91aa76.png)

___

babel doesn't have any way to set a config from the CLI, but it does have `--no-babelrc`. allowed config names are:
![image](https://user-images.githubusercontent.com/4071474/68508602-a5faa800-0223-11ea-9c57-2a2f23fc4ef1.png)

# thoughts

sorted by order of how strongly i feel

## `--rc-file` should be `--config`

It seems to just be a config, so shouldn't it be called that? I didn't find any tools that name their config option like this.

## remove ability to customize config path

I think babel does this to simplify compiling across project boundaries. That doesn't apply much here, but I wonder if there is any purpose to have multiple config files for LHCI? We could consider removing any option to configure, and add it in later if demand is there. If we do this, I'd recommend the file be called `.lhciconfig.json` (or `.lhcirc.json`)

## allow `.js`

right now just `.json` is allowed. this makes sense for now, and we should definitely defer this. just calling it out now

## configure via package.json

ditto above section

## nit on `rc`

`rc` stands for run commands*, and so I'd think files called `rc` should be something you must run (shell, node, etc). I wouldn't consider a json file to be a executable file. I see that eslint uses `eslintrc.js` and `eslintrc.json` (ditto for babel), so that doesn't jive with my understanding, and it's a popular enough tool that I concede any argument for using `.lhcirc.js` and `.lhciconfig.json`.

\* at least, it did initially. it has morphed into `run configuration` over time, which of course kills my argument. but I did some light research on this and can't find a satisfactory source on this definition. from what I currently know, I consider this morph to be a mistake.

patrickhulce:
## Bold Highlighted Proposal

Let's name this file `.lighthouserc`. Everything is already nested under the `"ci"` property in preparation for this exact conversation and other lighthouse settings could slide into here nicely.

Everyone with me? :D

Alternative File Names

- `lighthouseci.json`, simple, straight forward
- `lhci.json`, simple, straight forward, short
- `lhci.config.json`, simple, straight forward
- `.lighthousecirc`, lighthouse circ works I guess
- `.lhcirc`, acronym hell

[...]

PR #104, opened 2019-11-18
fix (cli): use rc-file if it wasn't found recursively

denar90:
> One more time without issue

So, since we are kinda has advanced usage in Action, `rc-file` has custom name like `lighthouserc_assertions.json` and fails when I upgrated to v0.3.3 - https://github.com/treosh/lighthouse-ci-action/pull/15/checks?check_run_id=307103933

It happens because file can't be found recursively via using[ from cli ](https://github.com/GoogleChrome/lighthouse-ci/blob/v0.3.3/packages/cli/src/cli.js#L36)calling [findRcFile](https://github.com/GoogleChrome/lighthouse-ci/blob/v0.3.3/packages/utils/src/lighthouserc.js#L71), then using `assertCmd.runCommand` it fails because parsed config was not passed.

I'm not 100% it's a correct fix. Here are 2 options for the fix:
1. One in PR
2. Use kinda the same logic as it's used in [autorunCmd.runCommand](https://github.com/GoogleChrome/lighthouse-ci/blob/v0.3.3/packages/cli/src/autorun/autorun.js#L99) for [assertCmd.runCommand](https://github.com/GoogleChrome/lighthouse-ci/blob/v0.3.3/packages/cli/src/assert/assert.js#L44) to get `rc-file` values.

Cheers,

### DIFF
--- a/packages/cli/src/cli.js
+++ b/packages/cli/src/cli.js
@@ -29,13 +29,11 @@ updateNotifier({pkg}).notify({defer: false});
 /** @return {[string, (path: string) => LHCI.YargsOptions]|[LHCI.YargsOptions]} */
 function createYargsConfigArguments() {
   const simpleArgv = yargsParser(process.argv.slice(2), {envPrefix: 'LHCI'});
-  /** @type {[string, (path: string) => LHCI.YargsOptions]} */
-  const configOption = ['config', loadAndParseRcFile];
   // If they're using the config option or opting out of auto-detection, use the config option.
-  if (simpleArgv.config || hasOptedOutOfRcDetection()) return configOption;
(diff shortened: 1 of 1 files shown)


denar90:
CI with YARN_GPG job failed with `The command "choco install googlechrome" failed and exited with 127 during .` :(

patrickhulce:
I think there might be some confusion @denar90, there is no more option named `rc-file` or `rcFile` anymore, it was renamed before 0.3.0 during the alpha release to `config` (see discussion in #48) 

CI action needs to update  https://github.com/treosh/lighthouse-ci-action/blob/64530652461989e1cb2e537e8dd13bf6476cdf83/src/index.js#L26 to use `--config`

denar90:
Yeah, my bad. @patrickhulce thank you)
 
> Guy got too excited to contribute one more time 🤣

PR #125, opened 2019-12-01
docs(cli): add yarn installation example

var-bin:
Hi,

I added configs for [yarn](https://yarnpkg.com)

Cheers 👋,
Vitalii

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -8,6 +8,12 @@ The CLI is the primary API you'll use when setting up Lighthouse CI. Install the
 npm install -g @lhci/cli
 ```
 
+or
+
+```bash
+yarn global add @lhci/cli
+```
+
 ## Commands
 
 All commands support configuration via a JSON file passed in via `--config=./path/to/`. Any argument on the CLI can also be passed in via environment variable. For example, `--config=foo` can be replaced with `LH_RC_FILE=foo`. Learn more about [configuration](./configuration.md).


patrickhulce:
Thank you very much for the contribution @var-bin! 😃 

As for the change itself, I'm not sure it's worth adding duplicate copies of documentation just to replace `npm` with `yarn`. IMHO, yarn users understand how to use yarn when shown npm commands and I'm concerned it might add some unnecessary complexity and increase the maintenance burden of updating documentation.

WDYT about just making the one-time reference to `yarn global add` in the CLI docs and once at beginning of getting started with a link on how to use yarn?

var-bin:
Hi @patrickhulce,

Thank you for your answer. From some point of view, I agree with you. The main reason why I want to ad duplicates is to be able to do copy-paste without any additional stuff. I agree that yarn users aware of the right commands but again, in this way, Lighthouse CI docs are going to be not so suitable.

> WDYT about just making the one-time reference to yarn global add in the CLI docs and once at beginning of getting started with a link on how to use yarn?

Do you mean to add a command to this doc without other changes? 
https://github.com/var-bin/lighthouse-ci/blob/master/docs/cli.md

patrickhulce:
Ah if this is just about copy-paste convenience then we definitely shouldn't create duplicate copies of all the documentation, copy-paste plus find-replace is easy enough :)

And yes, I'm proposing keeping your existing change to docs/cli.md and just one line in docs/getting-started.md to reference yarn docs.

var-bin:
Well, as I said before, I partially agree with you ;-). The main reason why I'd created this PR is I spent around 40 minutes and four updates in my branch for being able to use `yarn` instead of `npm`. I think I would spend about 5 minutes and only one update in the branch for setting the right CI with the right config in this repo. I believe this will help community in the long-term run.

Again, this is under discussion. I'm OK with adding changes only in two files without duplications that we were discussing before ;-)

patrickhulce:
I understand your point @var-bin, but I think the future maintenance burden as we expand the possible configurations is too high to commit to duplication with the current resources we have. Let's go with the changes in only two files for now.

I really appreciate your input and helpfulness though! :D

var-bin:
@patrickhulce I've updated current PR. The only file that I left with changes is `docs/cli.md`.

PR #136, opened 2019-12-03
Feat/111 add yml support config file

### LINKED ISSUE #111: request: yml support for config file (lighthouserc.json)
is this already supported or can support be added?

jzabala:
Add yaml support for config file.

Closes #111


docs/configuration.md
```diff
@@ -33,7 +37,9 @@ lhci assert --preset=lighthouse:recommended --assertions.uses-webp-images=off
 
 ## Structure
 
-The structure of a `lighthouserc.json` is segmented by command. Any options you see for a particular command in the [CLI documentation](./cli.md) can be set by the property of the same name in the config file.
+The structure of the config file is segmented by command. Any options you see for a particular command in the [CLI documentation](./cli.md) can be set by the property of the same name in the config file.
+
+##### `lighthouserc.json`
```
patrickhulce:
```suggestion
**`lighthouserc.json`:**
```

we've mostly been going with bold regular text for filenames within a section in the docs

docs/configuration.md
```diff
@@ -54,12 +60,31 @@ The structure of a `lighthouserc.json` is segmented by command. Any options you
 }
 ```
 
+##### `lighthouserc.yml`
```
patrickhulce: similar to the discussion in #125, I'm not sure we need to duplicate *all* of the documentation just to show how yaml works, yaml users should be familiar with how yaml translates to JSON and vice-versa :)

let's just leave a single `lighthouse.yml` example config in the `## Common Examples` section?

packages/utils/src/lighthouserc.js
```diff
@@ -42,17 +48,34 @@ function loadAndParseRcFile(pathToRcFile) {
...
+  // Convert all `.` in key names to `:`
   recursivelyReplaceDotInKeyName(rc);
   return rc;
 }
 
+/**
+ * Parse file content to JSON.
+ * @param {string} pathToRcFile
+ * @param {string} contents
+ * @return {Object}
```
patrickhulce:
```suggestion
 * @return {LHCI.LighthouseRc}
```

packages/utils/test/lighthouserc.test.js
```diff
@@ -13,23 +13,30 @@ const rc = require('../src/lighthouserc.js');
 const {safeDeleteFile} = require('../../cli/test/test-utils.js');
 
 describe('lighthouserc.js', () => {
+  function writeFile(path, content) {
```
patrickhulce: seems like this just passes the args directly to writeFileSync ;)

packages/utils/test/lighthouserc.test.js
```diff
@@ -13,23 +13,30 @@ const rc = require('../src/lighthouserc.js');
...
+    fs.writeFileSync(path, content);
+  }
   describe('#loadAndParseRcFile', () => {
-    let tmpFile;
-    function writeFile(json) {
-      fs.writeFileSync(tmpFile, JSON.stringify(json));
+    function writeJSONFile(path, json) {
+      writeFile(path, JSON.stringify(json));
```
patrickhulce:
```suggestion
      fs.writeFileSync(path, JSON.stringify(json));
```

packages/utils/test/lighthouserc.test.js
```diff
@@ -13,23 +13,30 @@ const rc = require('../src/lighthouserc.js');
...
 
 describe('lighthouserc.js', () => {
+  function writeFile(path, content) {
+    fs.writeFileSync(path, content);
+  }
   describe('#loadAndParseRcFile', () => {
-    let tmpFile;
-    function writeFile(json) {
-      fs.writeFileSync(tmpFile, JSON.stringify(json));
+    function writeJSONFile(path, json) {
```
patrickhulce: let's use a `writeYAMLFile` helper too that uses `yaml.safeDump` instead of `JSON.stringify` so we can share the test definitions between them :)

packages/utils/test/lighthouserc.test.js
```diff
@@ -38,40 +45,138 @@ describe('lighthouserc.js', () => {
...
-      const expected = path.join(LH_ROOT, 'lighthouserc.json');
-      expect(rc.findRcFile(LH_ROOT)).toEqual(expected);
+    describe('should find an rcfile in the directory', () => {
+      it('.json', () => {
+        const expected = path.join(LH_ROOT, 'lighthouserc.json');
+        expect(rc.findRcFile(LH_ROOT)).toEqual(expected);
+      });
+      it('.yaml', async () => {
+        const expected = path.join(FIXTURES_DIR, 'lighthouserc.yaml');
+        writeFile(expected, '');
```
patrickhulce: can we use a tmp directory for these if we're writing to them in tests? don't want to worry about commiting them later

docs/configuration.md
jzabala: Done!

docs/configuration.md
jzabala: Sure. I am fine with one example. 

I added the: **YAML Advanced config** at the end. I think it gives a good example about how to use the yaml config file.

packages/utils/test/lighthouserc.test.js
jzabala: 🤦🏼‍♂️🤦🏼‍♂️🤦🏼‍♂️ what a silly mistake 😋.

Fixed.

packages/utils/test/lighthouserc.test.js
jzabala: Done.

packages/utils/test/lighthouserc.test.js
jzabala: Done.

packages/utils/test/lighthouserc.test.js
jzabala: Great suggestion.

Created a temp dir for all files.

jzabala:
> Awesome job on the implementation @jzabala! Mostly just docs and tests comments :)
> 
> thanks so much for taking this on! 👏👏

Thank you very much @patrickhulce for taking the time to review my PR. 

I addressed all the comments.

Ready to go 😁

packages/utils/src/lighthouserc.js
jzabala: Done.

jzabala:
> Excellent work @jzabala and thanks for the super fast turnaround! LGTM 🎉

Glad I can help 🙂

packages/utils/src/lighthouserc.js
```diff
@@ -42,17 +48,34 @@ function loadAndParseRcFile(pathToRcFile) {
...
   return rc;
 }
 
+/**
+ * Parse file content to JSON.
+ * @param {string} pathToRcFile
+ * @param {string} contents
+ * @return {LHCI.LighthouseRc}
+ */
+function parseFileContentToJSON(pathToRcFile, contents) {
```
chris-kruining: If I may be nit-picky, there isn't any parsing to json, there is parsing from json or yaml. so I suggest to simply rename the method to `parseContent` and changing all the comments to reflect this

jzabala: No, you are not being nit-picky, it is great that you care about the name of things. Let me justify why I did choose the name. When content is passed to this function, it is one of two things: 1) a string yaml representation or 2) a string json representation. And the calls to yaml.safeLoad(contents) or JSON.parse(contents), is what parses the string to a javascript object. A more appropriate name would be `parseFileContentToJavaScriptObject` or just `parseFileContentToObject`, but since most devs know what JSON is and how it looks by calling it `parseFileContentToJSON`, I know devs will immediately imagine that the return of this function is something like: `{...}`. The problem that I see with `parseContent` is that by looking at the name I don't know what would be the result of this function. Will it be parsing content to `string`, `number`, `boolean`? Not sure by the name.

chris-kruining: I see, and mostly agree, but by the same logic I'd say ToJSON is misleading, because I wont get json as result from the method, you get an object. So by that logic I propose `parseContentToObject` or perharps `parseConfigFromContent`(would be really stretching it)

jzabala: Yeah, a `ToObject` would be the most appropriate. Or maybe `ToPOJO` jejeje. I don't like a lot the `parseConfigFromContent` because the function is doing nothing in regards of checking that the content is a valid config. Content could be `{ 'foo": "bla" }` and it would still parse that without problem.

Issue #134, opened 2019-12-03
Support custom "Listening" pattern for startServerCommand


harmenjanssen:
Running lhci using a php local server does not seem to work: Lighthouse isn't picking up the server output.

With the following `.lighthouserc.json`: 

```json
{
  "ci": {
    "collect": {
      "numberOfRuns": 3,
      "startServerCommand": "php -S localhost:60000 ./server.php",
      "url": [
        "http://localhost:60000/nl/styleguide"
      ]
    }
  }
}
```

I get the following error:

```
$ lhci autorun
✅  .lighthouseci/ directory writable
✅  Configuration file found
✅  Chrome installation found
Healthcheck passed!

Started a web server with "php -S localhost:60000 ./server.php"...
WARNING: Timed out waiting for the server to start listening.
         Ensure the server prints "listening" when it is ready.
```

Even though the output from the server contains the word "listening":

```
$ php -S localhost:60000 ./server.php
PHP 7.3.7 Development Server started at Tue Dec  3 09:54:41 2019
Listening on http://localhost:60000
Document root is /Users/harmen/Sites/foo
Press Ctrl-C to quit.
```

Not sure who's at fault here, if I can debug this further please let me know, I'd like to help fix this.

patrickhulce:
Just to be clear, the whole process doesn't work or just the timeout message happens erroneously? I'm not able to reproduce the server not working, but I do get the unexpected warning message.

harmenjanssen:
You're right, the server is working while Lighthouse is running. 
I do get a lot of warnings in the report, like the following snippet:

```
"runWarnings": [
    "Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests. (Status code: 500)"
  ],
  "runtimeError": {
    "code": "ERRORED_DOCUMENT_REQUEST",
    "message": "Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests. (Status code: 500)"
  }
```

But to be honest I'm very new to this and piecing it together, so that might as well be caused by something else.

harmenjanssen:
Alright, I've created a reduced testcase and the runWarnings I was getting in my previous comment were my own fault.

So, to summarize, the issue is this: Lighthouse is erroneously timing out and showing an error, but is working fine otherwise.

patrickhulce:
[...]

PR #140, opened 2019-12-05
feat(collect): support custom listening pattern for startServerCommand

jzabala:
This PR introduces the changes to support custom listening pattern for `startServerCommand` through the new collect option `startServerReadyPattern`.

Closes https://github.com/GoogleChrome/lighthouse-ci/issues/134

### DIFF
--- a/packages/cli/src/collect/collect.js
+++ b/packages/cli/src/collect/collect.js
@@ -45,6 +45,11 @@ function buildCommand(yargs) {
     startServerCommand: {
       description: 'The command to run to start the server.',
     },
+    startServerReadyPattern: {
+      description: 'String pattern to listen for started server.',
+      type: 'string',
+      default: 'listen|ready',
+    },
     settings: {description: 'The Lighthouse settings and flags to use when collecting'},
     numberOfRuns: {
       alias: 'n',
@@ -95,9 +100,10 @@ async function startServerAndDetermineUrls(options) {
   if (!options.staticDistDir) {
     let close = async () => undefined;
     if (options.startServerCommand) {
+      const regexPattern = new RegExp(options.startServerReadyPattern, 'i');
       const {child, patternMatch, stdout, stderr} = await runCommandAndWaitForPattern(
         options.startServerCommand,
-        /(listen|ready)/i,
+        regexPattern,
         {timeout: 10000}
       );
       process.stdout.write(`Started a web server with "${options.startServerCommand}"...\n`);
@@ -105,7 +111,9 @@ async function startServerAndDetermineUrls(options) {
 
       if (!patternMatch) {
         process.stdout.write(`WARNING: Timed out waiting for the server to start listening.\n`);
-        process.stdout.write(`         Ensure the server prints "listening" when it is ready.\n`);
+        // This is only for readability.
+        const message = `Ensure the server prints a matching pattern ${regexPattern} when it is ready.\n`;
+        process.stdout.write(`         ${message}`);
         if (process.env.CI) process.stderr.write(`\nServer Output:\n${stdout}\n${stderr}\n`);
       }
     }
--- a/packages/cli/test/cli.test.js
+++ b/packages/cli/test/cli.test.js
@@ -9,8 +9,8 @@
 
 jest.retryTimes(3);
 
-const os = require('os');
 const path = require('path');
+const os = require('os');
 const {spawn} = require('child_process');
 const fetch = require('isomorphic-fetch');
 const log = require('lighthouse-logger');
@@ -29,7 +29,6 @@ describe('Lighthouse CI CLI', () => {
   const rcMatrixFile = path.join(__dirname, 'fixtures/lighthouserc-matrix.json');
   const rcExtendedFile = path.join(__dirname, 'fixtures/lighthouserc-extended.json');
   const budgetsFile = path.join(__dirname, 'fixtures/budgets.json');
-  const staticDistDir = path.join(__dirname, 'fixtures');
   const tmpSqlFilePath = getSqlFilePath();
 
   let server;
@@ -148,30 +147,8 @@ describe('Lighthouse CI CLI', () => {
     });
   });
 
+  // FIXME: Tests dependency. Moving these tests breaks others.
   describe('collect', () => {
-    it('should collect results from staticDistDir', () => {
-      const {stdout, stderr, status} = runCLI([
-        'collect',
-        `--config=${rcFile}`,
-        `--static-dist-dir=${staticDistDir}`,
-      ]);
-
-      const stdoutClean = stdout;
-      expect(stdoutClean).toMatchInlineSnapshot(`
-        "Started a web server on port XXXX...
-        Running Lighthouse 2 time(s) on http://localhost:XXXX/checkout.html
-        Run #1...done.
-        Run #2...done.
-        Running Lighthouse 2 time(s) on http://localhost:XXXX/index.html
-        Run #1...done.
-        Run #2...done.
-        Done running Lighthouse!
-        "
-      `);
-      expect(stderr.toString()).toMatchInlineSnapshot(`""`);
-      expect(status).toEqual(0);
-    }, 90000);
-
     it('should collect results with a server command', () => {
       // FIXME: for some inexplicable reason this test cannot pass in Travis Windows
       if (os.platform() === 'win32') return;
@@ -196,7 +173,6 @@ describe('Lighthouse CI CLI', () => {
       expect(stderr.toString()).toMatchInlineSnapshot(`""`);
       expect(status).toEqual(0);
     }, 60000);
-
     it('should collect results from explicit urls', () => {
       const {stdout, stderr, status} = runCLI([
         'collect',
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -110,25 +110,27 @@ lhci collect --help
 Run Lighthouse and save the results to a local folder
 
 Options:
-  --help                   Show help                                      [boolean]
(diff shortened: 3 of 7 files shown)


jzabala:
🤔 This is weird. A bunch of things seems to be failing in the ci, but all the tests pass locally for me.

packages/cli/src/collect/collect.js
```diff
@@ -97,15 +102,19 @@ async function startServerAndDetermineUrls(options) {
...
+        new RegExp(options.startServerReadyPattern, 'i'),
         {timeout: 10000}
       );
       process.stdout.write(`Started a web server with "${options.startServerCommand}"...\n`);
       close = () => killProcessTree(child.pid);
 
       if (!patternMatch) {
         process.stdout.write(`WARNING: Timed out waiting for the server to start listening.\n`);
-        process.stdout.write(`         Ensure the server prints "listening" when it is ready.\n`);
+        process.stdout.write(
```
patrickhulce: this got a tad hard to tell why the indent is there :)

how about `const pattern = new RegExp(options.startServerReadyPattern, 'i')` at the top of this block and then

```suggestion
        const msg = `Ensure the server prints text matching ${pattern} when it's ready.\n`;
        process.stdout.write(`         ${message}`);
```

?

packages/cli/test/cli.test.js
```diff
@@ -197,6 +197,32 @@ describe('Lighthouse CI CLI', () => {
       expect(status).toEqual(0);
     }, 60000);
 
+    it('should collect results with a server command with custom start pattern', () => {
```
patrickhulce: would you mind moving this test to its own test file? maybe `collect-start-server.test.js`?

cli.test.js is by far the longest running bottleneck so would like to avoid increasing its duration if possible :)

packages/cli/src/collect/collect.js
jzabala: Sure, but do you want to print the whole regex pattern or just the pattern? The indentation was done by prettier because the line was too long after I added `options.startServerReadyPattern`.

The whole regex pattern would look something like this: `/listen|ready/i` and just the pattern `listen|ready`.

packages/cli/test/cli.test.js
jzabala: how about if I move all collect related tests to their own file, something like `ci-collect.test.js`?  I think this way will scale better because later we can move the others to: `ci-upload.test.js`, `ci-sever.test.js`, and so on. 

Or maybe `ci.collect.test.js`

jzabala:
> looks perfect as always, many thanks @jzabala! just a few nits :)

Thanks @patrickhulce. I appreciate your comments 🙂

packages/cli/src/collect/collect.js
patrickhulce: regex.source works for me too, though I don't have a strong preference 🤷‍♂  

I thought I recognized the prettier autoflow 😉 I was trying to think of something that preserves just the indentation piece on one line to make it obvious with the previous `process.stdout.write`

packages/cli/test/cli.test.js
patrickhulce: let's just move this one out for now, I agree cli.test.js will need to be broken apart eventually but let's leave that refactoring for another day. so far I've just been trying to add all *new* tests into their own file

jzabala: Hey @patrickhulce sorry, I am just watching your comment but I already moved one test to `ci-collect.js`, and added one for the timeout failing case. If you prefer me to change it back, I can do it, no problem. I just went ahead and implemented it after leaving you the first comment. Bad strategy on my part 😅.

packages/cli/src/collect/collect.js
jzabala: I went ahead with the full regex option. Maybe it can help someone debugging a start command not working 🤷🏼‍♂️.

packages/cli/test/cli.test.js
```diff
@@ -187,16 +164,15 @@ describe('Lighthouse CI CLI', () => {
 
       const stdoutClean = stdout.replace(/sqlDatabasePath=.*?"/, 'sqlDatabasePath=<file>"');
       expect(stdoutClean).toMatchInlineSnapshot(`
-        "Started a web server with \\"yarn start server -p=XXXX --storage.sqlDatabasePath=<file>\\"...
```
patrickhulce: revert these few lines? :)

packages/cli/test/ci-collect.test.js
```diff
@@ -0,0 +1,117 @@
+/**
+ * @license Copyright 2019 Google Inc. All Rights Reserved.
```
patrickhulce: let's just name this file `collect.test.js`

packages/cli/test/ci-collect.test.js
```diff
@@ -0,0 +1,117 @@
...
+const fs = require('fs');
+const {runCLI} = require('./test-utils.js');
+
+describe('collect', () => {
+  const fixturesDir = path.join(__dirname, 'fixtures');
+  const rcFile = path.join(fixturesDir, 'lighthouserc.json');
+
+  let tempDir;
+  beforeAll(() => {
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'lighthouserc-'));
```
patrickhulce: FYI I added a `withTmpDir` command to utils that auto cleans up everything

https://github.com/GoogleChrome/lighthouse-ci/blob/b26fe670994d81ab3a5931bb2bcc98cc25f0f207/packages/cli/test/test-utils.js#L47-L51

example usage:

```js
it('should test', async () => {
  await withTmpDir(dir => {
    // do something with the dir
  })
})
```

packages/cli/test/ci-collect.test.js
```diff
@@ -0,0 +1,117 @@
...
+
+    const serverPath = path.join(fixturesDir, 'autorun-start-server/autorun-server.js');
+    const startCommand = `SERVER_START_MESSAGE='Running server' node ${serverPath}`;
+    const {stdout, status} = runCLI(
+      [
+        'collect',
+        `-n=1`,
+        `--config=${rcFile}`,
+        `--start-server-command=${startCommand}`,
+        '--url=http://localhost:52425/',
```
patrickhulce: these have the same port that seems like a problem, maybe make it another env variable?

packages/cli/src/collect/collect.js
```diff
@@ -95,17 +100,20 @@ async function startServerAndDetermineUrls(options) {
   if (!options.staticDistDir) {
     let close = async () => undefined;
     if (options.startServerCommand) {
+      const regexPattern = new RegExp(options.startServerReadyPattern, 'i');
```
akanshgulati: We should honour the default regex `/(listen|ready)/i` if no specific regex is provided. Am I missing anything?

patrickhulce: yep it's specified as the default value in yargs config above @akanshgulati

jzabala: The default regex is set as the `startServerReadyPattern` default.

packages/cli/test/ci-collect.test.js
jzabala: 🤔 why is that a problem? From what I see the server is stopped after the test or we would be getting a _something is already running in that port_ error.

patrickhulce: when everything goes as expected it shouldn't :) but we've seen flaky tests in lighthouse-core when some race means the port isn't released back into the os pool on CI machines fast enough and so the next one will flakily fail, just easier to use unique ports, plus it means we can run these in parallel in the future since we're already doing some of the split work :)

jzabala: Got it. 

As the lib grows we should consider a solution like: https://www.npmjs.com/package/get-port. 

It can get hard to keep an eye on every port to make sure we are not repeating them.

patrickhulce: Ironically it's the exact same method that `get-port` uses that we've seen get us into trouble on super slow machines (opening up and quickly closing a server on a port) 😆but point taken, yes a util for generating unique ports would be helpful

packages/cli/test/cli.test.js
jzabala: Prettier not helping me after moving tests around 😅

jzabala: Done!

packages/cli/test/ci-collect.test.js
jzabala: Done.

packages/cli/test/ci-collect.test.js
jzabala: Cool, using it now `withTmpDir`. 

By the way we should `try... finally` that function. If the test fails the dir won't be deleted.

packages/cli/test/ci-collect.test.js
jzabala: 🤣 what a small world.

packages/cli/test/ci-collect.test.js
patrickhulce: sg! the cleanup isn't *super* critical since it's in a tmpdir anyhow but definitely a great point if it's worth that ugly indentation it's worth a try finally :)

patrickhulce:
great job @jzabala, many thanks! 🎉

harmenjanssen:
Awesome! Thanks @jzabala, this helps my case a lot. 🙂

jzabala:
> Awesome! Thanks @jzabala, this helps my case a lot. 🙂

Glad I can help @harmenjanssen 👍🏼

PR #148, opened 2019-12-10
feat(cli): Add warning when using chrome-flags and puppeteerScript

### LINKED ISSUE #143: Add warning when using chrome-flags and puppeteerScript
I had confused about chrome flags settings.

In my project, using custom userAgent. So first, I tried these settings.

```json
{
  "ci": {
    "collect": {
      "settings": {
        "chrome-flags": "--user-agent=something"
      }
    }
  }
}
```

It only worked before I add `puppeteerScript` options.
If the user using `puppeteerScript`, lighthouse's chrome flags options will be ignored and the settings should be changed like this.

```json
{
  "ci": {
    "collect": {
      "puppeteerScript": "script.js",
      "puppeteerLaunchOptions": {
        "args": ["--user-agent=something"]
      }
    }
  }
}
```

It confused me.
I suggest these two options.
1. Re-position chrome-flags option like this.
```json
{
  "ci": {
    "collect": {
      "chromeFlags": ["--user-agent=something"],
      "puppeteerScript": "script.js",
    }
  }
}
```

2. Just show warnings when that case.
```shell
lhci collect
[WARN] Your chrome flags options will be ignored because of... blah blah
```

What's your opinion?
Can I handle this issue?

mysticPrg:
- add checkIgnoredChromeFlagsOption function
- add test code for checkIgnoredChromeFlagsOption function

closes #143

### DIFF
--- a/packages/cli/src/collect/collect.js
+++ b/packages/cli/src/collect/collect.js
@@ -145,11 +145,29 @@ async function startServerAndDetermineUrls(options) {
 
 /**
  * @param {LHCI.CollectCommand.Options} options
+ * @return {void}
+ */
+function checkIgnoredChromeFlagsOption(options) {
+  const usePuppeteerScript = !!options.puppeteerScript;
+  const useChromeFlags = options.settings ? !!options.settings.chromeFlags : false;
+
+  if (usePuppeteerScript && useChromeFlags) {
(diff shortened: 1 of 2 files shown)


mysticPrg:
@googlebot I fixed it.

Issue #145, opened 2019-12-09
Show LHCI server version on UI


ravick12:
@patrickhulce 
     Is there currently a way to find out which version the LHCI server is running?  I guess I can look at the package.json in the docker container to see which version of @lhci/server is running. But it will be good to have some place in UI (e.g footer or in the side menu), that shows the version number.

patrickhulce:
@ravick12 you can hit `/version` URL on the server to find out what version it's running. Adding it to the sidebar UI isn't a bad idea though :)

ravick12:
@patrickhulce thanks for the version link.

patrickhulce:
This was fixed by #150 🎉

wodka:
:/ it always shows up as 0.1.0 -> the package.json version is not up to date

`/version` returns 0.3.12

patrickhulce:
Ah thanks @wodka the build process probably changed since #150 that broke this, we'll use the API version instead 👍

PR #150, opened 2019-12-11
Show package version in server UI sidebar

gravi2:
@patrickhulce I am raising this PR to address #145. 

Below is a quick overview of what has changed:
1.  I am making use of process.env.npm_package_version to make this change at the build time. I assume the `lerna version` lifecycle scripts will update the package.json  and the same will be picked by `parcel` in the `build` command.
2. JSX Changes in page-sidebar.jsx to add the siderbar footer and version elements. 
3. CSS changes in app.css and page-sidebar.css

Attaching few screenshots for your reference.

desktop

mobile

### DIFF
--- a/packages/server/src/ui/app.css
+++ b/packages/server/src/ui/app.css
@@ -97,6 +97,8 @@
 
   --page-sidebar-width: 260px;
   --page-sidebar-nav-item-height: 24px;
+  --page-sidebar-footer-font-size: 14px;
+  --page-sidebar-footer-margin: 25px;
 
   --gauge-size: 82px;
   --gauge-stroke-width: 7px;
--- a/packages/server/src/ui/app.css
+++ b/packages/server/src/ui/app.css
@@ -97,8 +97,6 @@
 
   --page-sidebar-width: 260px;
   --page-sidebar-nav-item-height: 24px;
-  --page-sidebar-footer-font-size: 14px;
-  --page-sidebar-footer-margin: 25px;
 
   --gauge-size: 82px;
   --gauge-stroke-width: 7px;
--- a/packages/server/src/ui/layout/page-sidebar.css
+++ b/packages/server/src/ui/layout/page-sidebar.css
@@ -71,6 +71,12 @@
 }
 
 .page-sidebar__footer {
+  position: absolute;
+  left: 0;
+  right: 0;
+  bottom: var(--base-spacing);
+  text-align: center;
+}
   padding: 0 var(--page-header-horizontal-padding);
   width: 100%;
   display: flex;
@@ -81,5 +87,5 @@
 }
 
 .page-sidebar__version {
-  color:sandybrown;
+  color: var(--inverted-secondary-text-color);
 }
--- a/packages/server/src/ui/layout/page-sidebar.css
+++ b/packages/server/src/ui/layout/page-sidebar.css
@@ -77,14 +77,6 @@
   bottom: var(--base-spacing);
   text-align: center;
 }
-  padding: 0 var(--page-header-horizontal-padding);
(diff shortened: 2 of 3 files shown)


packages/server/src/ui/app.css
```diff
@@ -97,6 +97,8 @@
 
   --page-sidebar-width: 260px;
   --page-sidebar-nav-item-height: 24px;
+  --page-sidebar-footer-font-size: 14px;
```
patrickhulce: let's not use a different font-size for this, how about `--subtext-font-size`?

packages/server/src/ui/app.css
```diff
@@ -97,6 +97,8 @@
...
+  --page-sidebar-footer-margin: 25px;
```
patrickhulce: we try not to use bespoke spacing, we have `var(--base-spacing)` for margin and padding :)

packages/server/src/ui/layout/page-sidebar.jsx
```diff
@@ -51,6 +51,9 @@ export const PageSidebar = props => {
           }}
         />
       </div>
+      <div className="page-sidebar__footer">
+        <div className="page-sidebar__version">Version {process.env.npm_package_version}</div>
```
patrickhulce:
```suggestion
        <div className="page-sidebar__version">
          {/* This environment variable is set by npm/yarn on any script command. */}
          v{process.env.npm_package_version}
         </div>
```

packages/server/src/ui/layout/page-sidebar.css
```diff
@@ -69,3 +69,17 @@
...
+  width: 100%;
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  margin: var(--page-sidebar-footer-margin) 0;
+  font-size: var(--page-sidebar-footer-font-size);
+}
+
+.page-sidebar__version {
+  color:sandybrown;
```
patrickhulce:
```suggestion
  color: var(--inverted-secondary-text-color);
```

packages/server/src/ui/layout/page-sidebar.css
```diff
@@ -69,3 +69,17 @@
   color: var(--inverted-text-color);
   text-decoration: none;
 }
+
+.page-sidebar__footer {
```
patrickhulce:
```suggestion
.page-sidebar__footer {
  position: absolute;
  left: 0;
  right: 0;
  bottom: var(--base-spacing);
  text-align: center;
}
```

maybe something like that?

packages/server/src/ui/app.css
gravi2: @patrickhulce I have accepted your CSS changes and I have removed --page-sidebar-footer-font-size as we don't need it any more.

packages/server/src/ui/app.css
gravi2: @patrickhulce I have accepted your CSS changes and I have removed --page-sidebar-footer-maring as we don't need it any more.

gravi2:
@patrickhulce Thanks for your review comments. I have addressed the review feedback and the same is now available in the PR. Here is the latest screenshot after review changes, for your reference. 
Screenshot from 2019-12-12 01-22-30

gravi2:
@patrickhulce I have fixed the linting issue using the autofix (--fix). Though I din't quite like the lint fix it did :)

patrickhulce:
>  Though I din't quite like the lint fix it did :)

Haha yeah that one doesn't make the most sense to me either, looks like there's an open bug in Prettier for it already. (https://github.com/prettier/prettier/issues/7132#issuecomment-565489232)


Thanks @gravi2!

Issue #157, opened 2019-12-17
Add SPA support to fallback server


lili21:
the fallback-server does not support front-end router right now.

patrickhulce:
Thanks for filing @lili21, good idea!

This should be fairly straightforward  if anyone is looking for their first contribution :) 

Basically add 
```js
app.get('*', (_, res) => res.sendFile(path.join(pathToBuildDir, 'index.html')))
```
to
https://github.com/GoogleChrome/lighthouse-ci/blob/3a34a7b0a9f57b78ec1d5c1be5263262408c61fa/packages/cli/src/collect/fallback-server.js#L18-L24

and a test to go along with it

Amagon96:
Working on it! n.n/

rdlabo:
I found this Issue after making pull request👀
https://github.com/GoogleChrome/lighthouse-ci/pull/159

@Amagon96 How is the progress? If you haven't started yet, I'd be happy to get feedback on this code.

Thanks.

Amagon96:
Hi @rdlabo  actually I already done, I'll be requesting the PR soon since I just woke up :p

PR #159, opened 2019-12-19
feat(collect): Make it easy to use with SPA

rdlabo:
Now we can't set both `startServerCommand` and `staticDistDir`. 

ref: https://github.com/GoogleChrome/lighthouse-ci/blob/master/packages/cli/src/collect/collect.js#L100-L102

So server return 404 not found at sub-page in SPA.

I think `isSinglePageApplication` flag is very useful. (When using no SPA, 404 information is important)

Thanks.

closes https://github.com/GoogleChrome/lighthouse-ci/issues/157

### DIFF
--- a/packages/cli/src/collect/fallback-server.js
+++ b/packages/cli/src/collect/fallback-server.js
@@ -13,12 +13,16 @@ const createHttpServer = require('http').createServer;
 class FallbackServer {
   /**
    * @param {string} pathToBuildDir
+   * @param {boolean|undefined} isSinglePageApplication
    */
-  constructor(pathToBuildDir) {
+  constructor(pathToBuildDir, isSinglePageApplication) {
     this._pathToBuildDir = pathToBuildDir;
     this._app = express();
     this._app.use(compression());
     this._app.use('/', express.static(pathToBuildDir));
+    if (isSinglePageApplication === true) {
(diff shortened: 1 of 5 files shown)


rdlabo:
related: https://github.com/GoogleChrome/lighthouse-ci/issues/157

packages/cli/src/collect/fallback-server.js
```diff
@@ -13,12 +13,16 @@ const createHttpServer = require('http').createServer;
...
    * @param {string} pathToBuildDir
+   * @param {boolean|undefined} isSinglePageApplication
    */
-  constructor(pathToBuildDir) {
+  constructor(pathToBuildDir, isSinglePageApplication) {
     this._pathToBuildDir = pathToBuildDir;
     this._app = express();
     this._app.use(compression());
     this._app.use('/', express.static(pathToBuildDir));
+    if (isSinglePageApplication === true) {
```
Amagon96: I think this can be simplified to 
`if (isSinglePageApplication)`

rdlabo: Thanks for review. I fixed!

rdlabo:
closed https://github.com/GoogleChrome/lighthouse-ci/issues/157

rdlabo:
@patrickhulce Thanks for review. I added simple test case for `isSinglePageApplication` flag. 
Thanks!

PR #165, opened 2020-01-01
Update README.md

otherwise777:
added docker-compose example and clarified initial run

### DIFF
--- a/docs/recipes/docker-server/README.md
+++ b/docs/recipes/docker-server/README.md
@@ -64,5 +64,17 @@ kubectl get service
 # kubernetes    ClusterIP      10.X.X.X       <none>         443/TCP        9m
 # lhci-server   LoadBalancer   10.X.X.X       X.X.X.X        80:XXXXX/TCP   2m
 ```
+### Docker-compose
+
+An example docker-compose.yml could look like this:
+```  lhci:
+    container_name: lhci_server
+    image: patrickhulce/lhci-server
+    volumes:
+      - lhci-data:/data
+    ports:
+      - 9001:9001```
+
 
 Note that the server has no authentication mechanisms and that anyone with HTTP access to the server will be able to view and create data. If your server contains sensitive information, consider protecting it within a firewall that is only accessible to your internal network.
+To setup the server you can run ```lhci wizard``` locally (with lhci installed) and make sure you have your 'serverBaseUrl' in your 'lighthouserc.json' configured


otherwise777:
@googlebot I signed it!

docs/recipes/docker-server/README.md
```diff
@@ -64,5 +64,17 @@ kubectl get service
 # kubernetes    ClusterIP      10.X.X.X       <none>         443/TCP        9m
 # lhci-server   LoadBalancer   10.X.X.X       X.X.X.X        80:XXXXX/TCP   2m
 ```
+### Docker-compose
+
+An example docker-compose.yml could look like this:
+```  lhci:
```
patrickhulce: is this a working docker-compose.yml? I think we're missing some parts like `version: `, `services: `, etc

docs/recipes/docker-server/README.md
```diff
@@ -64,5 +64,17 @@ kubectl get service
 # kubernetes    ClusterIP      10.X.X.X       <none>         443/TCP        9m
 # lhci-server   LoadBalancer   10.X.X.X       X.X.X.X        80:XXXXX/TCP   2m
 ```
+### Docker-compose
+
+An example docker-compose.yml could look like this:
```
patrickhulce: an explanation of why you'd want to use docker-compose for this would be great. 

AFAICT, there's no real benefit to using it because the lhci server is just a single docker container and there's not really much to orchestrate. `docker container run --publish 9001:9001 --mount='source=lhci-data,target=/data' --detach lhci` accomplishes the same task, no?

docs/recipes/docker-server/README.md
```diff
@@ -64,5 +64,17 @@ kubectl get service
...
+    container_name: lhci_server
+    image: patrickhulce/lhci-server
+    volumes:
+      - lhci-data:/data
+    ports:
+      - 9001:9001```
+
 
 Note that the server has no authentication mechanisms and that anyone with HTTP access to the server will be able to view and create data. If your server contains sensitive information, consider protecting it within a firewall that is only accessible to your internal network.
+To setup the server you can run ```lhci wizard``` locally (with lhci installed) and make sure you have your 'serverBaseUrl' in your 'lighthouserc.json' configured
```
patrickhulce: maybe we move this and the auth notes up right below `Deployment` and edit this to read
```suggestion
Refer to the [Getting Started documentation](../../getting-started.md#the-lighthouse-ci-server) for how to configure Lighthouse CI to connect to your new server.
```

docs/recipes/docker-server/README.md
otherwise777: This is working in my setup, with the version do you mean the tag? That's in the image name. The service is not required for this container

docs/recipes/docker-server/README.md
otherwise777: Good idea, as long as it's more clear that there is a step required after deployment

docs/recipes/docker-server/README.md
otherwise777: Depends in yoyr setup, but indeed there's no benefit for it, the docker run and docker compose do the same thing. I use docker compose to keep all my running containers in the same file and not having to run a bunch of docker runs on startup.

patrickhulce:
Hey @otherwise777 sorry for lagging here, but another contributor put up a slightly different take that ended up landing (#218) so I think we're going to close this. Hopefully the information you wanted to see made it all in and feel free to open a new PR if you still feel something is missing! :)

Issue #127, opened 2019-12-02
Add mysql support


lili21:
looks like the server only support postgresql ?

patrickhulce:
Correct, onl

patrickhulce:
this was fixed in #175 by @HappyTobi! 🎉 awesome work

PR #175, opened 2020-01-09
NEW add mysql support

HappyTobi:
Hi all,

I added the mysql support to the lighthouse-ci server.

I changed the following things:
- add mysql driver.
- update documentation (cli.md).
- fix the issue with the options.limit value in combination with mysql.
- added test for mysql (like postgress test) - travis should be changed to run that test.

Related to issue: #127

Kind regards
Tobi


HappyTobi:
@googlebot I signed it!

packages/server/src/api/storage/sql/sql.js
```diff
@@ -79,6 +79,7 @@ function createSequelize(options) {
   const sequelizeOptions = {
     operatorsAliases: false,
     logging: () => {},
+    type: Sequelize.QueryTypes.BULKUPDATE,
```
patrickhulce: what does this do? I can't find `type` in the constructor options https://sequelize.org/master/class/lib/sequelize.js~Sequelize.html#instance-constructor-constructor

packages/server/src/api/storage/sql/sql.js
```diff
@@ -298,7 +299,7 @@ class SqlStorageMethod {
     const builds = await this._findAll(buildModel, {
       where: {projectId, ...omit(options, ['limit'], {dropUndefined: true})},
       order,
-      limit: options.limit || 10,
+      limit: Number(options.limit) || 10,
```
patrickhulce: hm this is a bug elsewhere if this isn't `number|undefined` already...

I'll fix that another time, good catch!

packages/server/test/mysql-server.test.js
```diff
@@ -0,0 +1,47 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+/* eslint-env jest */
+
+const runTests = require('./server-test-suite.js').runTests;
+const runServer = require('../src/server.js').createServer;
+
+describe('postgres server', () => {
```
patrickhulce:
```suggestion
describe('mysql server', () => {
```

packages/server/test/mysql-server.test.js
```diff
@@ -0,0 +1,47 @@
...
+  if (!process.env.MYSQL_DB_URL) {
+    it.skip('should work on postgres', () => {});
```
patrickhulce:
```suggestion
    it.skip('should work on mysql', () => {});
```

yarn.lock
```diff
@@ -5,7 +5,6 @@
 "3d-view@^2.0.0":
   version "2.0.0"
   resolved "https://registry.yarnpkg.com/3d-view/-/3d-view-2.0.0.tgz#831ae942d7508c50801e3e06fafe1e8c574e17be"
-  integrity sha1-gxrpQtdQjFCAHj4G+v4ejFdOF74=
```
patrickhulce: please revert these changes, and we should have mysql and new sequelize updates in the yarn.lock

deleting integrity might mean your yarn is out of date?

HappyTobi:
Thx!

Yep I will do that later.

Kind regards

packages/server/test/mysql-server.test.js
HappyTobi: thx, changed

packages/server/test/mysql-server.test.js
HappyTobi: thx, changed

packages/server/src/api/storage/sql/sql.js
HappyTobi: So the problem was that you can pass only numbers (isNaN is always false) but when you look into the sequelize logs you see that the query looks like
`SELECT * ... LIMIT '100'.`
But that didn't work in mysql it has to be a number like
`SELECT * ... LIMIT 100`
and wir the explicit cast it works correct

patrickhulce: Oh I know the reason for the fix! :)

Just saying that the typedef expresses `number|undefined` as the type within this function, so this cast shouldn't be *here*. Instead it should be in projects.js where we blindly pass `req.query` which has the limit as the incorrect string type.

yarn.lock
HappyTobi: thx for that information ... and yes my yarn installation was super old.
I will update the yarn.lock file with the Integrity checksum.

packages/server/src/api/storage/sql/sql.js
HappyTobi: I will change it

packages/server/src/api/storage/sql/sql.js
HappyTobi: This is an option for the queries that will made with sequelize: [QueryTypes](https://sequelize.org/v4/variable/index.html#static-variable-QueryTypes)

Without the option it's not possible to run all the migrations with "umzug" because there you made a bulk update. 

But i read the code, documentation again and I found out that it's better to set the "queryType" to the bulk options in the migration scripts itself. 

The type will dropped out of the sql.js file and added to the bulk operations direct

docs/cli.md
```diff
@@ -288,7 +288,7 @@ Options:
   --logLevel        [string] [choices: "silent", "verbose"] [default: "verbose"]
   --port, -p                                            [number] [default: 9001]
   --storage.sqlDialect
-                    [string] [choices: "sqlite", "postgres"] [default: "sqlite"]
+                    [string] [choices: "sqlite", "postgres","mysql"] [default: "sqlite"]
```
patrickhulce:
```suggestion
                    [string] [choices: "sqlite", "postgres", "mysql"] [default: "sqlite"]
```

packages/server/src/api/routes/projects.js
```diff
@@ -69,6 +69,9 @@ function createRouter(context) {
   router.get(
     '/:projectId/builds',
     handleAsyncError(async (req, res) => {
+      if (!isNaN(req.query.limit)) {
+        req.query.limit = parseInt(req.query.limit);
```
patrickhulce: 👍

packages/server/src/api/routes/projects.js
```diff
@@ -69,6 +69,9 @@ function createRouter(context) {
   router.get(
     '/:projectId/builds',
     handleAsyncError(async (req, res) => {
+      if (!isNaN(req.query.limit)) {
```
patrickhulce: nit: the coercion behavior of isNaN and parseInt are different and somewhat confusing, how about 

```suggestion
      if (Number.isInteger(parseInt(req.query.limit))) {
```

packages/server/src/api/storage/sql/build-model.js
```diff
@@ -24,9 +24,11 @@ const attributes = {
   avatarUrl: {type: Sequelize.STRING(256)},
   ancestorHash: {type: Sequelize.STRING(40)},
   externalBuildUrl: {type: Sequelize.STRING(256)},
-  runAt: {type: Sequelize.DATE()}, // should mostly be equal to createdAt but modifiable by the consumer
-  committedAt: {type: Sequelize.DATE()},
-  ancestorCommittedAt: {type: Sequelize.DATE()},
+  runAt: {type: Sequelize.DATE(6)}, // should mostly be equal to createdAt but modifiable by the consumer
```
patrickhulce: these are all because of https://dev.mysql.com/doc/refman/8.0/en/fractional-seconds.html, right?

> The fsp value, if given, must be in the range 0 to 6. A value of 0 signifies that there is no fractional part. If omitted, the default precision is 0. (This differs from the standard SQL default of 6, for compatibility with previous MySQL versions.)

packages/server/src/api/storage/sql/migrations/20191009-project-token.js
```diff
@@ -14,7 +14,12 @@ module.exports = {
    */
   up: async (queryInterface, Sequelize) => {
     await queryInterface.addColumn('projects', 'token', {type: Sequelize.UUID()});
-    await queryInterface.bulkUpdate('projects', {token: Sequelize.col('id')}, {token: null});
+    await queryInterface.bulkUpdate(
+      'projects',
+      {token: Sequelize.col('id')},
+      {token: null},
+      {type: Sequelize.QueryTypes.BULKUPDATE}
```
patrickhulce: super weird that sequelize makes you do this when the name of the method is already bulkUpdate...

HappyTobi:
Yep I will do that, fine for me

packages/server/src/api/storage/sql/build-model.js
HappyTobi: Yes else mysql will not save any milliseconds

packages/server/src/api/storage/sql/migrations/20191009-project-token.js
HappyTobi: Yes, i read for a while because without the type you get an error. 
I think it was fixed in Sequelize V5

packages/server/src/api/routes/projects.js
HappyTobi: Yes that's better, because negotiation is not so easy to read.

packages/server/src/api/storage/sql/build-model.js
HappyTobi: You can also see a super short comment on the "Data types" at the Sequelize documentation:
https://sequelize.org/v4/manual/tutorial/models-definition.html
(But yes the mysql site gives more information about the problem)

packages/utils/test/api-client.test.js
```diff
@@ -51,7 +51,9 @@ describe('Lighthouse CI API Client', () => {
       await client.getBuilds('124', {branch: 'master'});
       expect(fetchMock).toHaveBeenCalledWith(
         `http://localhost:9000/v1/projects/124/builds?branch=master`,
-        {headers: {}}
+        {
```
patrickhulce: revert this one?

packages/cli/src/collect/collect.js
```diff
@@ -108,7 +108,9 @@ async function startServerAndDetermineUrls(options) {
       const {child, patternMatch, stdout, stderr} = await runCommandAndWaitForPattern(
         options.startServerCommand,
         regexPattern,
-        {timeout: 10000}
+        {
```
patrickhulce: revert this one?

packages/utils/test/api-client.test.js
HappyTobi: This happen with linter (—fix) but es I can revert that

packages/cli/src/collect/collect.js
HappyTobi: This happen with linter (—fix) but es I can revert that

patrickhulce:
Incredible job @HappyTobi thank you again!

patrickhulce:
published in 0.3.8! 🎉

PR #177, opened 2020-01-13
feat: add semantic-pull-request support

vinayakkulkarni:
- [x] Add support for [Semantic Pull Requests](https://probot.github.io/apps/semantic-pull-requests/)

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,5 @@
+# Always validate the PR title AND all the commits
+titleAndCommits: true
+# Allows use of Merge commits (eg on github: "Merge branch 'master' into feature/ride-unicorns")
+# this is only relevant when using commitsOnly: true (or titleAndCommits: true)
+allowMergeCommits: true
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -7,6 +7,3 @@ scopes:
   - utils
   - viewer
   - recipes
-# Allows use of Merge commits (eg on github: "Merge branch 'master' into feature/ride-unicorns")
-# this is only relevant when using commitsOnly: true (or titleAndCommits: true)
-allowMergeCommits: false
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -2,4 +2,4 @@
 titleAndCommits: false
 # Allows use of Merge commits (eg on github: "Merge branch 'master' into feature/ride-unicorns")
 # this is only relevant when using commitsOnly: true (or titleAndCommits: true)
-allowMergeCommits: true
+allowMergeCommits: false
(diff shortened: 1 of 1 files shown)


vinayakkulkarni:
@googlebot I signed it!

.github/semantic.yml
```diff
@@ -0,0 +1,5 @@
+# Always validate the PR title AND all the commits
+titleAndCommits: true
+# Allows use of Merge commits (eg on github: "Merge branch 'master' into feature/ride-unicorns")
+# this is only relevant when using commitsOnly: true (or titleAndCommits: true)
+allowMergeCommits: true
```
patrickhulce: let's set this to `false`, we only use rebase or squash and merge

.github/semantic.yml
```diff
@@ -0,0 +1,5 @@
+# Always validate the PR title AND all the commits
+titleAndCommits: true
```
patrickhulce: let's do `titleOnly`, it seems less invasive to newcomers who will typically end up with squashed commits anyhow

vinayakkulkarni: - [x] Fixed in 48a65795c00df13544f0a0d60fc75c4fcb0c9f6c

.github/semantic.yml
vinayakkulkarni: - [x] Fixed in 612db406731da7c58de5bed89c0a931f1de52355

.github/semantic.yml
```diff
@@ -0,0 +1,14 @@
+# Always validate the PR title, and ignore the commits
+titleOnly: true
+# Always validate the PR title AND all the commits
```
patrickhulce: Oh, I guess we don't need these two lines if we're already using `titleOnly: true` correct?

.github/semantic.yml
```diff
@@ -0,0 +1,14 @@
...
+titleAndCommits: false
+# You can define a list of valid scopes
+scopes:
+  - cli
+  - server
+  - utils
+  - viewer
+  - recipes
+# Allows use of Merge commits (eg on github: "Merge branch 'master' into feature/ride-unicorns")
+# this is only relevant when using commitsOnly: true (or titleAndCommits: true)
```
patrickhulce: does this mean we don't need this flag at all?

.github/semantic.yml
vinayakkulkarni: According to the [configuration](https://github.com/zeke/semantic-pull-requests#configuration) docs, all these settings are optional.
I personally prefer explicitly setting all the configuration settings.

I'll remove:
```diff
- titleAndCommits: false
```

vinayakkulkarni: - [x] Fixed in 21b0609

.github/semantic.yml
vinayakkulkarni: Yep, we don't need it.

- [x] Fixed in 80d2a34

PR #186, opened 2020-01-17
Allow use of merge commits in build hash; add Gitlab CI variable to get hash

jimvandervoort:
I noticed lhci uses the last commit _that is not a merge commit_ when I ran in Gitlab CI. I think this behaviour is not desired because most CI providers set variables like `CIRCLE_SHA1` to the actual commit (even if it is a merge commit), so this creates inconsistencies between CI providers. This PR fixes two things:

- Read `CI_COMMIT_SHA` env variable to get the commit hash from Gitlab CI.
- When falling back to getting the hash form `git`, do not exclude merge commit so the hash is equal to that of any env variables. I also think this is better because people can still break something when resolving merge conflicts.

### DIFF
--- a/packages/utils/src/build-context.js
+++ b/packages/utils/src/build-context.js
@@ -67,7 +67,7 @@ function getCurrentHash() {
   ]);
   if (envHash) return envHash;
 
-  const result = childProcess.spawnSync('git', ['rev-parse', 'HEAD'], {
+  const result = childProcess.spawnSync('git', ['rev-list', '--no-merges', '-n1', 'HEAD'], {
     encoding: 'utf8',
   });
   if (result.status !== 0) {
--- a/packages/utils/src/build-context.js
+++ b/packages/utils/src/build-context.js
@@ -67,7 +67,7 @@ function getCurrentHash() {
   ]);
   if (envHash) return envHash;
 
-  const result = childProcess.spawnSync('git', ['rev-list', '--no-merges', '-n1', 'HEAD'], {
+  const result = childProcess.spawnSync('git', ['rev-parse', 'HEAD'], {
     encoding: 'utf8',
   });
   if (result.status !== 0) {
--- a/packages/utils/src/build-context.js
+++ b/packages/utils/src/build-context.js
@@ -62,6 +62,8 @@ function getCurrentHash() {
     'TRAVIS_COMMIT',
     // Circle CI
     'CIRCLE_SHA1',
+    // GitLab CI
+    'CI_COMMIT_SHA',
   ]);
   if (envHash) return envHash;
 


patrickhulce:
Thanks for the contribution! The `--no-merges` behavior is intentional for several reasons:

1. Several CI providers automatically merge master into the branch before building which means if we considered merges the hash that we report would be one that doesn't exist anywhere else and is relatively meaningless, we only want to report a hash that will exist for the user elsewhere.
1. Avoid 100% of commit message explanations for builds being `Merge X into Y` which anecdotally isn't as helpful compared to a real commit message (though certainly depends on whether your branch naming conventions are better than your commit message conventions 😆 )
1. > set variables like CIRCLE_SHA1 to the actual commit (even if it is a merge commit)
This behavior is OK and avoids the dangers of our biggest reason for doing this (see 1). I've yet to see a CI provider set their variable to a transient hash that only exists in their build system.
1. This is just a default fallback, so we'd rather be on the safe side for 1 and 2 and if you are doing something outside of our supported systems, we've [documented](https://github.com/GoogleChrome/lighthouse-ci/blob/2204f91151a2558a1ea3fe68677c5f6a9385747f/docs/cli.md#build-context) and encouraged folks to inform LHCI explicitly of your build context.

As we expand coverage of the CI env variables (thanks for the Gitlab bit!) this should all become less of a problem too. 

WDYT?

jimvandervoort:
@patrickhulce I did not consider that some CI indeed auto-merge master, that's def. a reason to keep it. Considering that LHCI intents to support all CI systems I think you're right and it's better to keep using `--no-merges` as fallback.

I reverted my commit, one thing I'd like to note is that the behaviour did change for Gitlab CI right now: considering a fairly default CI setup (no detached pipelines), a report is generated for the `Merge branch X into Y` commit, where previously it would report the commit from the branch `X`. This is the behaviour that I intended, but maybe we should document that somewhere?

patrickhulce:
> a report is generated for the Merge branch X into Y commit, where previously it would report the commit from the branch X. This is the behaviour that I intended, but maybe we should document that somewhere?

Is the `Merge branch X into Y` commit that now gets shown for Gitlab CI a commit that you created as a result of you running `git merge X` and `git push`? Or is the `Merge branch X into Y` commit one that Gitlab CI generated to keep the branch up-to-date like in case 1 above?

If it's the former, then I think this is WAI and is consistent with other providers. If it's the latter then Gitlab CI is inconsistent and we should probably leave as-is.

jimvandervoort:
It's the former 🙂. The commit exists for everyone.

Alright then; I believe everything WAI and is consistent with other CI providers, I only meant to highlight that previously no merge commits would ever show up in LHCI for Gitlab.

I think we're good to go then?

patrickhulce:
Gotcha, yes I think we're good to go!

aruku:
I'm sorry to post on a closed thread, but this is the closest discussion to my question that I've been able to find: so the _upload_ command uses the hash and the message from the last non-merge commit even though the audit is run on a merge commit? Wouldn't this confuse people?

aruku:
Also, is there a way to prevent _--no-merges_? Can't find it on the docs.

jimvandervoort:
@aruku a quick peek in master suggests me that `--no-merges` is no longer used:

https://github.com/jimvandervoort/lighthouse-ci/blob/master/packages/utils/src/build-context.js#L91-L93

When I made this PR that command was:

https://github.com/jimvandervoort/lighthouse-ci/blob/master/packages/utils/src/build-context.js#L91-L93

jimvandervoort:
Nevermind, that actually finds the commit time, not the commit hash, `--no-merges` is still used in `getCurrentHash`. I don't see a way to configure that either atm.

PR #200, opened 2020-01-29
NEW add configuration argument to wizard command

HappyTobi:
Hi,

i created the pr because we're running the `lhci server` behind a reverse proxy with authentication.
Now it's not possible to the the `lhci wizard` on a local machine because the wizard can't reach the server (missing authentication headers).

To fix that issue I changed the wizard command that a user can pass the `--config` argument (like autorun)

Example: 
`lhci wizard --config=./lighthouserc.json`

In the passed configuration it was now possible to set two different things:
- extraHeaders
- serverBaseUrl

The extraHeaders will be uses when the `lhci server` will be called.  
The serverBaseUrl can be used to set a default location for the remote `lhci server` address. 

Example configuration file: 
```javascript
{
  "ci": {
  "wizard": {
      "serverBaseUrl": "https://github.com/GoogleChrome/lighthouse-ci",
      "extraHeaders":" {\"Authorization\": \"Basic <BASE64>"}"
    }
  }
}
```

Hope you like the implementation. 
I think to pass a configuration is better than changing the wizard for the user because passing an argument was a known concept in using `lhci`

Kind regards
Tobi

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -291,7 +291,7 @@ module.exports = {
 
 ### Custom headers for running the wizard
 
-If you're running the `lhci server` behind an reverse proxy or any other component that requires some `extra headers` you can configure them in the wizard section `extraHeaders`.
+If you're running the `lhci server` behind a reverse proxy or any other component that requires some extra headers you can configure them in the wizard section `extraHeaders`.
 
 ```json
 {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -308,7 +308,7 @@ If you're running the `lhci server` behind an reverse proxy or any other compone
 
 ### Default lhci server location for running the wizard
 
-If you're running the `lhci wizard` multiple times, you can configure a default `serverBaseUrl` that you have to type them in at each `lhci wizard` run.
+If you're running the `lhci wizard` multiple times, you can configure a default `serverBaseUrl` to avoid typing it in at each `lhci wizard` run.
 
 ```json
 {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -58,6 +58,9 @@ module.exports = {
     server: {
       // server options here
     },
+    wizard: {
+      // wizard options here
+    }
   },
 };
 ```
@@ -78,6 +81,9 @@ module.exports = {
     },
     "server": {
       // server options here
+    },
+    "wizard": {
+      // wizard options here
     }
   }
 }
@@ -98,6 +104,9 @@ ci:
 
   server:
     # server options here
+  
+  wizard:
+    # wizard options here
 ```
 
 ## Recommendations
@@ -279,6 +288,39 @@ module.exports = {
   },
 };
 ```
+
+### Custom headers for running the wizard
+
+If you're running the `lhci server` behind an reverse proxy or any other component that requires some `extra headers` you can configure them in the wizard section `extraHeaders`.
+
+```json
+{
+  "ci": {
+    "wizzard": {
+      "extraHeaders": "{\"Authorization\": \"Basic content\"}"
+    }
+  }
+}
+```
+
+> **_NOTE:_** The `wizard` options will be overriten by the `upload` options, to be sure that the wizard options will be used, create a seperate config file
+
+
+### Default lhci server location for running the wizard
+
+If you're running the `lhci wizard` multiple times, you can configure a default `serverBaseUrl` that you have to type them in at each `lhci wizard` run.
+
+```json
+{
+  "ci": {
+    "wizzard": {
+      "serverBaseUrl": "https://localhost:3000/",
+    }
+  }
+}
+```
+
+> **_NOTE:_** The `wizard` options will be overriten by the `upload` options, to be sure that the wizard options will be used, create a seperate config file
 
 ### YAML Advanced config
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -303,7 +303,7 @@ If you're running the `lhci server` behind a reverse proxy or any other componen
 }
 ```
 
-> **_NOTE:_** The `wizard` options will be overriten by the `upload` options, to be sure that the wizard options will be used, create a seperate config file
+> **_NOTE:_** The `wizard` options will be overwritten by the `upload` options, to be sure that the wizard options will be used, create a separate config file.
 
 
 ### Default lhci server location for running the wizard
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -320,7 +320,7 @@ If you're running the `lhci wizard` multiple times, you can configure a default 
 }
 ```
 
-> **_NOTE:_** The `wizard` options will be overriten by the `upload` options, to be sure that the wizard options will be used, create a seperate config file
+> **_NOTE:_** The `wizard` options will be overwritten by the `upload` options, to be sure that the wizard options will be used, create a separate config file.
 
 ### YAML Advanced config
 
--- a/packages/cli/src/wizard/wizard.js
+++ b/packages/cli/src/wizard/wizard.js
@@ -50,7 +50,10 @@ async function runNewProjectWizard(options) {
     },
   ]);
 
-  const api = new ApiClient({rootURL: responses.serverBaseUrl || options.serverBaseUrl, extraHeaders: options.extraHeaders || {}});
+  const api = new ApiClient({
+    rootURL: responses.serverBaseUrl || options.serverBaseUrl,
+    extraHeaders: options.extraHeaders || {},
+  });
   const project = await api.createProject({
     name: responses.projectName,
     externalUrl: responses.projectExternalUrl,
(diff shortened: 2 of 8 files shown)


packages/cli/src/wizard/wizard.js
```diff
@@ -7,13 +7,21 @@
...
+const _ = require('@lhci/utils/src/lodash.js');
 const log = require('lighthouse-logger');
 
 /**
  * @param {import('yargs').Argv} yargs
  */
 function buildCommand(yargs) {
-  return yargs;
+  return yargs.options({
+    config: {description: 'The lighthouserc.json file preferences.'},
```
patrickhulce: this option should already be set in `src/cli.js` for all commands, so shouldn't be necessary here.

packages/cli/src/wizard/wizard.js
```diff
@@ -60,6 +70,14 @@ async function runNewProjectWizard(options) {
  * @return {Promise<void>}
  */
 async function runCommand(options) {
+  const rcFilePath = resolveRcFilePath(options.config);
```
patrickhulce: you shouldn't need to manually do any of this either, just change 

https://github.com/GoogleChrome/lighthouse-ci/blob/2fe16195dab6a8e84995bdc66397d5efc39a4917/packages/utils/src/lighthouserc.js#L138

to do `...ci.wizard` at the beginning

packages/cli/src/wizard/wizard.js
```diff
@@ -26,8 +34,7 @@ async function runNewProjectWizard(options) {
       type: 'input',
       name: 'serverBaseUrl',
       message: 'What is the URL of your LHCI server?',
-      when: () => !options.serverBaseUrl,
```
patrickhulce: why present the prompt when we already have it configured?

packages/cli/test/fixtures/lighthouserc.json
```diff
@@ -22,6 +22,10 @@
       "storage": {
         "sqlDatabasePath": "cli-test-fixtures.tmp.sql"
       }
+    },
+    "wizard": {
+      "serverBaseUrl": "https://github.com/GoogleChrome/lighthouse-ci",
```
patrickhulce: this is going to be the first collision of property names in the CI file between commands in the config given that they should be identical between them I don't think it's a big deal, but we should track it and try to resolve in next breaking change anyhow

packages/cli/test/cli.test.js
```diff
@@ -98,6 +98,28 @@ describe('Lighthouse CI CLI', () => {
...
+
+      await waitForCondition(() => wizardProcess.stdoutMemory.includes('Which wizard'));
+      await writeAllInputs(wizardProcess, [
+        '', // Just ENTER key to select "new-project"
+        '', // Just ENTER key to use serverBaseUrl from config file
+        'AwesomeCIProjectName', // Project name
+        'https://example.com', // External build URL
+      ]);
+
+      expect(wizardProcess.stdoutMemory).toContain('https://github.com/GoogleChrome/lighthouse-ci');
```
patrickhulce: I don't really understand these expectations, don't we want to test that the correct `serverBaseUrl` was hit?

WDYT about changing the test above to write to a tmp file that uses `http://localhost:${server.port}` as the `serverBaseUrl`?

packages/cli/src/wizard/wizard.js
HappyTobi: Thank yes its not needed i change that back to return yargs;

packages/cli/src/wizard/wizard.js
HappyTobi: Thx i didn't see that.
Because of the collision of `wizard.serverBaseUrl` and  `upload.serverBaseUrl` should i rename the option or just add something to the documentation that the user know that he has to define the `serverBaseUrl` only once? (at upload or wizard)?

packages/cli/src/wizard/wizard.js
HappyTobi: I remove that, to give the user feedback what kind of url was used.
So the User see what happened and can press enter to accept that.
If that's not the way you like I can bring the `when` check back.

packages/cli/test/fixtures/lighthouserc.json
HappyTobi: Thx, yes I think it should be documented or changed (wizard configuration part).
See my comment above.

packages/cli/src/wizard/wizard.js
patrickhulce: Alright works for me 👍

packages/cli/src/wizard/wizard.js
patrickhulce: Let's just add something to the documentation for now that the values can't be *different* but still tell folks set it in both places. In 0.4.0 I'm planning on restructuring the upload parameters to cluster by target type, they wouldn't collide in this case.

packages/cli/test/cli.test.js
patrickhulce: Hm, yeah I'm not sure that first case is worth testing given our current set of options. I'd vote to just modify the existing test to pass in `http://localhost:${server.port}` through a config.

types/wizard.d.ts
```diff
@@ -8,6 +8,8 @@ declare global {
   namespace LHCI {
     namespace WizardCommand {
       export interface Options {
+        config?: string;
```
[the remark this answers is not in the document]
HappyTobi: thx, done

packages/utils/src/lighthouserc.js
```diff
@@ -135,7 +135,7 @@ function hasOptedOutOfRcDetection(argv = process.argv, env = process.env) {
 function convertRcFileToYargsOptions(rcFile, pathToRcFile) {
   const ci = flattenRcToConfig(rcFile);
   /** @type {LHCI.YargsOptions} */
-  let merged = {...ci.assert, ...ci.collect, ...ci.upload, ...ci.server};
+  let merged = {...ci.wizard, ...ci.assert, ...ci.collect, ...ci.upload, ...ci.server};
```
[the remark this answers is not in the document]
HappyTobi: i added a test for loading the wizard configuration and the overlapping configuration of the upload and wizard configuration.

HappyTobi:
@patrickhulce documentation are also added now. (configuration.md)

docs/configuration.md
```diff
@@ -280,6 +289,39 @@ module.exports = {
...
+{
+  "ci": {
+    "wizzard": {
+      "extraHeaders": "{\"Authorization\": \"Basic content\"}"
+    }
+  }
+}
+```
+
+> **_NOTE:_** The `wizard` options will be overriten by the `upload` options, to be sure that the wizard options will be used, create a seperate config file
```
patrickhulce:
```suggestion
> **_NOTE:_** The `wizard` options will be overwritten by the `upload` options, to be sure that the wizard options will be used, create a separate config file.
```

docs/configuration.md
```diff
@@ -280,6 +289,39 @@ module.exports = {
 };
 ```
 
+### Custom headers for running the wizard
+
+If you're running the `lhci server` behind an reverse proxy or any other component that requires some `extra headers` you can configure them in the wizard section `extraHeaders`.
```
patrickhulce:
```suggestion
If you're running the `lhci server` behind a reverse proxy or any other component that requires some extra headers you can configure them in the wizard section `extraHeaders`.
```

docs/configuration.md
```diff
@@ -280,6 +289,39 @@ module.exports = {
...
+  }
+}
+```
+
+> **_NOTE:_** The `wizard` options will be overriten by the `upload` options, to be sure that the wizard options will be used, create a seperate config file
+
+
+### Default lhci server location for running the wizard
+
+If you're running the `lhci wizard` multiple times, you can configure a default `serverBaseUrl` that you have to type them in at each `lhci wizard` run.
```
patrickhulce:
```suggestion
If you're running the `lhci wizard` multiple times, you can configure a default `serverBaseUrl` to avoid typing it in at each `lhci wizard` run.
```

docs/configuration.md
```diff
@@ -280,6 +289,39 @@ module.exports = {
...
+{
+  "ci": {
+    "wizzard": {
+      "serverBaseUrl": "https://localhost:3000/",
+    }
+  }
+}
+```
+
+> **_NOTE:_** The `wizard` options will be overriten by the `upload` options, to be sure that the wizard options will be used, create a seperate config file
```
patrickhulce:
```suggestion
> **_NOTE:_** The `wizard` options will be overwritten by the `upload` options, to be sure that the wizard options will be used, create a separate config file.
```

packages/cli/test/fixtures/lighthouserc.json
patrickhulce: bump :)

PR #218, opened 2020-02-18
docs(docker-server): adds docker compose example

wavemotionio:


wavemotionio:
@googlebot I signed it!

docs/recipes/docker-server/README.md
```diff
@@ -65,4 +65,21 @@ kubectl get service
 # lhci-server   LoadBalancer   10.X.X.X       X.X.X.X        80:XXXXX/TCP   2m
 ```
 
+### Docker Compose
+
+You can create a docker-compose.yml file. Execute `docker-compose up` relative to the file's location.
```
patrickhulce: an explanation of why you'd want to use docker-compose for this would be great.

AFAICT, there's no real benefit to using it on its own because the lhci server is just a single docker container and there's not really much to orchestrate. `docker container run --publish 9001:9001 --mount='source=lhci-data,target=/data' --detach lhci` accomplishes the same task, no?

wavemotionio: Admittedly, this approach doesn't have a huge benefit but it will save some people time:

1. Docker-compose should already be available to anyone using docker locally.
2. This will save users on Windows machines from having to slightly modify the repository to get it working (this is a two-step ordeal I can detail further if you'd like).
3. This snippet could be used verbatim in an Azure DevOps App Service or similar.

patrickhulce: Ah 2 and 3 are great reasons (would love to hear more about 2 separately as well!) mind adding 3 to the comment?

Maybe just...
```suggestion
You can create a docker-compose.yml file to orchestrate an LHCI server in cloud services or along with your existing app docker containers. Execute `docker-compose up` relative to the file's location.
```

patrickhulce:
btw the travis failure is a windows checksum flake that I just fixed in master so next push should fix it

docs/recipes/docker-server/README.md
wavemotionio: Yes, very nice suggestion!

wavemotionio:
@patrickhulce hey, to follow up on the conversation... it seemed to get this working on windows I had to modify https://github.com/GoogleChrome/lighthouse-ci/tree/master/docs/recipes/docker-server. Particularly, setting (lighthouserc.json) `"sqlDatabasePath": "/data/lhci.db"` to `"sqlDatabasePath": "lhci.db"` and `--mount='source=lhci-data,target=/data'` to `--mount='source=lhci-data,target=\data'` in the command line.

patrickhulce:
hm `"sqlDatabasePath": "lhci.db"` will use the current working directory (/usr/src/lhci) as the path to the database which isn't on the persistent data volume so I would expect the database to disappear. it's working as expected for you though?

those paths are inside the alpine-based docker container so the `/` *should* be working, I don't have much experience with docker on Windows though. this is after going through things [like this](https://rominirani.com/docker-on-windows-mounting-host-directories-d96f3f056a2c)?

PR #224, opened 2020-02-21
fix(assetions): normalize number assertion

BrunnerLivio:
When using a number as assertion in the lighthouse configuration, the `normalizeAssertion` function will return a faulty "normalized assertion".

Example of a JSON which would have failed before this PR:
```json 
{
  "ci": {
    "assert": {
      "assertions": {
        "total-byte-weight": "0.74",
      }
    }
  }
}
```

Note: Even though the assertion value is written as string, it will get resolved as a number. I do not know where this is coming from though (maybe `JSON.parse`?).

### DIFF
--- a/packages/utils/src/assertions.js
+++ b/packages/utils/src/assertions.js
@@ -62,7 +62,7 @@ const isFiniteNumber = x => typeof x === 'number' && Number.isFinite(x);
  */
 function normalizeAssertion(assertion) {
   if (!assertion) return ['off', {}];
-  if (typeof assertion === 'string') return [assertion, {}];
+  if (typeof assertion === 'string' || typeof assertion === 'number') return [assertion, {}];
   return assertion;
 }
 


patrickhulce:
Thanks for opening @BrunnerLivio! But the example you've provided is an invalid assertion format we don't support. In fact, we aren't intentionally trying to support anything even like

```js
{
  "assertions": {
    "total-byte-weight": 0.74
  }
}
```

what is that trying to assert? the numeric value? the score? The only reason `typeof string` is supported for normalization is for converting
```js
{
  "assertions": {
    "total-byte-weight": "error"
  }
}
```

into
```js
{
  "assertions": {
    "total-byte-weight": ["error", {}]
  }
}
```

BrunnerLivio:
@patrickhulce I see, my bad then! :)

patrickhulce:
No worries if there was something confusing in the docs that misled you, I'd love to fix that instead! :)

BrunnerLivio:
@patrickhulce I actually don’t recall why I have used that setting in the first place. I guess a `.lighthousec.json` intelli-sense or a JSON schema checker would have spared me from this submission - but that is just nagging on a high level :)

Issue #229, opened 2020-02-26
Lighthouse Status "details" link not pointing to where I expect


mansona:
Hi folks 👋 first of all, thanks for the awesome tool 🎉 it's exactly what I am looking for and has the perfect mix of complexity and simplicity for what I like 😂 

I've just set up the system for the first time on a project (see the PR here: https://github.com/ember-learn/ember-website/pull/607) but when I click the "details" link beside the Lighthouse Status Check I am redirected to https://github.com/GoogleChrome/lighthouse-ci instead of what I expected, the actual build itself 🤔 

Am I missing something because the build doesn't have a "base build" yet because it's the PR that is introducing the system? Or is this working as expected?

patrickhulce:
Thanks for filing @mansona! That's definitely unexpected it should be linking you to the comparison view on your LHCI server. I'll look into it ASAP :)

patrickhulce:
Turns out this was actually a server-side issue with the GitHub App, it's been restarted now and rerunning the build should fix this without any need to update `@lhci/cli` 👍

mansona:
wow thanks @patrickhulce that's a really quick turnaround on this 🎉 you're an absolute star 🌟

PR #230, opened 2020-02-26
Add GitHub App to Dogfood

patrickhulce:
trying to repro #229

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ jobs:
       os: windows
       before_install:
         - export LHCI_GITHUB_TOKEN="" # prevent any accidental leaking of the token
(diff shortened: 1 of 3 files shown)


abraham:
This change will cause repos to update their config if they are configured to require a green lhci run before merging.

Screen Shot 2020-03-19 at 20 49 13

patrickhulce:
ouch, that's an unfortunate side effect I didn't realize @abraham thanks for raising! I guess we'll go back for now and reintroduce this in 0.4.0 as a breaking change.

patrickhulce:
FYI @abraham 0.3.12 has been published that reverts this to just `lhci` 👍

Issue #179, opened 2020-01-13
Redesign of the project dashboard


patrickhulce:
Discrete components and steps:

## UI Components

- [x] Category score line chart
  - [x] basic line graph (8baaabb19afd8e39a75b3fdbf7ddb57d588e862d)
  - [x] add points (deaa0d0435741c8b8155cfbb8ebfb5bad25780a4)
  - [x] add y-axis label (8baaabb19afd8e39a75b3fdbf7ddb57d588e862d)
  - [x] add x-axis date range (9d22646730ab21b2a442ab1f3bbc783c55cd8a4d)
  - [x] colorize line graph based on score (b92ac77acf5354fd6dbafbd1b236ed91f6ae56a6)
    - [x] rework colorized approach based on masks (6e222ffe6aea9371b18ec3991521d1b7e6aac838)
  - [x] add # of builds picker (9d22646730ab21b2a442ab1f3bbc783c55cd8a4d)
  - [x] hover panel
    - [x] basic date and score on hover (35c366e41d1166bf03729daf1f9a2e70f48ea391)
    - [x] add category gauge (b304a30bbfa80b97b8e43e0c3705883c0c370098)
    - [x] add view report links (d164c400b393ed706330e1b5f39f8a8804a217ad)
    - [x] pin on click (6c038b9d65e07ba39acf12d3a391bbdec9a016c3)
  - [x] add vertical tracking line on hover or pin (fd4948a4e7e4c994297c12482d0d22418f7db771)
  - [x] add delta bar graph (4c315537f829a09747b81bbbde6d082e226551e5)
  - [x] add max/min gray error bars (08eeed993a6d2802e53e90ddfa70947cc305a75e)
- [x] url/branch picker (83e59e7e3691677fde4c6c8eafda89803f3935cd)
  - [x] move url/branch picker up to header on scroll (83e59e7e3691677fde4c6c8eafda89803f3935cd)
- [x] metric line chart
  - [x] basic line graph with 1 metric on fixed scale (112b204d6033dbd757429554fea4f2dda34ca296)
  - [x] axis markers (112b204d6033dbd757429554fea4f2dda34ca296)
  - [x] add other metrics by line type (f2fd3b3e5ebae11401027f29cfa173f29b49a5dd)
  - [x] dynamic scale (112b204d6033dbd757429554fea4f2dda34ca296)
  - [x] add colorization of line on hover (4a6c5eca609b4290852f050570177e9457162d56)
[...]

PR #233, opened 2020-03-02
Add distributions to server dashboard

patrickhulce:
image

ref #179


connorjclark:
any way we could flag the percentiles of interest here? (50, 75, 95, ..)

patrickhulce:
> any way we could flag the percentiles of interest here? (50, 75, 95, ..)
Maybe not so interesting for lab data as it would be for field ...

Yeah I'm not sure we want to. In lab we consider those percentiles mostly due to environmental variance and not really worthy of highlighting haha :)

patrickhulce:
Also the meaning of percentiles is kinda weird in this scenario since spread out over different builds over time. It's more like "this was the state of your app when it was worse than 75% of other examples we saw"

PR #267, opened 2020-04-09
fix(cli): tweak format of assertion results

paulirish:
The motivation is that when I had multiple URLs' results in the assertion dump...  I had a hard time figuring out if the "  16 result(s) for <url>" text was summarizing all the stuff above it, or providing a header of the stuff below it.

so i added a colon.

Also, I think formatting-wise it's a little clearer to indent the audittitle/docs. And i think we dont need the "documentation: " label.

---------

running the unit tests didnt really work for me.

I got a couple of these: 
image

and then many of the tests that were affected by these text changes werent run. i never saw failures. (i updated the snapshots by hand! 😨 )

i saw some other things too that I couldn't explain, but ... i'll leave it there for now. 

🤞 that the CI's tests run this with no problem! :)

### DIFF
--- a/packages/cli/src/assert/assert.js
+++ b/packages/cli/src/assert/assert.js
@@ -77,7 +77,7 @@ async function runCommand(options) {
       return levelA.localeCompare(levelB) || auditIdA.localeCompare(auditIdB);
     });
 
-    process.stderr.write(`${sortedResults.length} result(s) for ${log.bold}${url}${log.reset}\n`);
+    process.stderr.write(`${sortedResults.length} result(s) for ${log.bold}${url}${log.reset} :\n`);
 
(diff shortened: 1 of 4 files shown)


paulirish:
> 🤞 that the CI's tests run this with no problem! :)

suhweeeeeeeeeet! ✅

patrickhulce:
> running the unit tests didnt really work for me.
I got a couple of these:

Did you run the tests with `yarn test:unit`? Because of all the child processes that get spun up even my many core 2020 macbook tends to timeout a lot of the tests when run in max parallel (which is what happens if run with just `jest`), not sure what's up with that otherwise.

also possible there's something up with how fish handles the test scripts I didn't expect that messes with things :)

Issue #119, opened 2019-11-27
☂️👣 Breaking Changes for 0.4.0


patrickhulce:
A comprehensive list of changes that are being saved for a breaking 0.4.0 version.

## Last 0.3.x release

- [x] CLI healthcheck for hitting incompatible server
- [x] visual cue for version change in timeline
- [x] visual cue in build view for version change
- [x] reset project token wizard
- [x] fix issue with dashboard choosing bad URL by default

## Must Have

- [x] change default minScore to 0.9 https://github.com/GoogleChrome/lighthouse-ci/issues/295#issuecomment-622535215
- [x] move dom-size audit out of recommended preset
- [x] remove old dashboard design
- [x] upgrade to lighthouse 6.0 🎉 
  - [x] transferSize of budgets
  - [x] graphs for LCP & TBT
  - [x] support subRows
  - [x] support source-location details
  - [x] update presets with 6.0 audits
  - [x] upgrade the actual collection to 6.0
  - [ ] ???
- [x] replace `_average` statistics with `_median`
- [x] require project token for all created builds
- [x] recursive search and limit 5 with static-dist-dir by default (ref #95)

## Docs to update

- [x] 0.3.x references
- [x] timing budgets to add

## Not really breaking, for next release

- support a "root token" requirement to create new projects https://github.com/GoogleChrome/lighthouse-ci/issues/323
- PSI runner https://github.com/GoogleChrome/lighthouse-ci/issues/251
- PSI cron job support in server projects

## Punt

Churn without a ton of benefit.

- [ ] move upload options into objects nested by target type (for clarity and to resolve fact that wizard and upload have overlapping keys https://github.com/GoogleChrome/lighthouse-ci/pull/200#discussion_r372467370)

PR #271, opened 2020-04-15
feat(server): add warnings when versions change

patrickhulce:
build comparisons will appear in the usual runtime settings warning modal

dashboard will appear as shaded background on the time delta between the two builds where the change occurred and have a warning in the build overlay with the gauge

image

ref https://github.com/GoogleChrome/lighthouse-ci/issues/119


packages/server/src/api/statistic-definitions.js
```diff
@@ -9,6 +9,18 @@ const {computeRepresentativeRuns} = require('@lhci/utils/src/representative-runs
 
 /** @typedef {(lhrs: Array<LH.Result>) => ({value: number})} StatisticFn */
 
+/**
+ * @return {StatisticFn}
+ */
+function metaLighthouseVersion() {
+  return lhrs => {
+    const version = lhrs[0].lighthouseVersion || '';
+    const [_, major = '0', minor = '0', patch = '0'] = version.match(/^(\d+)\.(\d+)\.(\d+)/) || [];
```
connorjclark, line 18: do you care about the alpha/beta tag?

patrickhulce: not really. we'll never publish a version of lhci that's stuck on an alpha or beta tag and in the event that someone uploads data manually to lhci server using `lighthouse@next` I'm ok with the idea that they won't get warnings between alpha/beta/final versions

this behavior is also captured by the tests for this file so unlikely to accidentally change it either

patrickhulce:
super weird, this has already been merged not sure why it's open... 

https://github.com/GoogleChrome/lighthouse-ci/commit/8790c766c9c4a3692693c273c5d2668b88091f64

there's even activity from the pull bot saying it's been pulled into forks
image

PR #280, opened 2020-04-22
feat(cli): add user-timing assertion support

### LINKED ISSUE #279: Expose user-timings assertion
It would be nice to have access to user-timings data generated from the reports throught he `lhci` js API. It would make it a lot easier to closely monitor and track those if they are easily accessible through official APIs.

patrickhulce:
closes #279 

how does this API feel @WebCloud ?

### User Timings

The general format for asserting against a user timing value is `"user-timings:<kebab-cased-name>": ["<level>", {maxNumericValue: <value in milliseconds>}]`. For example, if you wanted to assert that a mark with name `My Custom Mark` started within the first 2s of page load and that a measure `my:custom-@-Measure` lasted fewer than 50 ms you would use the following assertions config.

Note that only the first matching entry with the name will be used from each run and the rest will be ignored.

```json
{
  "assertions": {
    "user-timings:my-custom-mark": ["warn", {"maxNumericValue": 2000}],
    "user-timings:my-custom-measure": ["error", {"maxNumericValue": 50}]
  }
}
```


WebCloud:
That looks great 🎉

PR #309, opened 2020-05-14
Update server to work with Lighthouse v6

patrickhulce:
ref #119 

- transferSize of budgets
- graphs for LCP & TBT
- handle subRow headings
- handle source-location details
- test cases for mixed v5/v6 data


connorjclark:
> handle subRow headings

FYI subrows are only used for experimental audits, so this could be done later.

patrickhulce:
> FYI subrows are only used for experimental audits, so this could be done later.

Yeah I know :) I just wanted to be prepared for the shape in case folks don't lock to our version of LH and end up putting 6.1.0 into a 0.4.0 server. Our tests used the --experimental preset 👍

PR #314, opened 2020-05-19
Assert shouldn't be nested under CI

skippednote:


skippednote:
@googlebot I signed it!

patrickhulce:
Thanks very much for the contribution @skippednote! It should be nested under `ci` though, it won't have any effect if it's just exported at the root.

The previous documentation lacked the prefix and just had a note that "all examples should be nested under `ci.assert` but will be omitted for brevity", and this just makes it clearer exactly how the file should look.

Are you seeing it work this way? Because that would be very unexpected :)

skippednote:
@patrickhulce I'm using `0.3.x` and it throws **No assertions to use** if `assert` is nested under `ci`.

skippednote:
```javascript
async function runCommand(options) {
  const {budgetsFile, assertions, assertMatrix, preset} = options;
```
Here the properties are not accessible as they are nested under `assert`

https://github.com/GoogleChrome/lighthouse-ci/blob/master/packages/cli/src/assert/assert.js#L49

patrickhulce:
I'm not able to reproduce this.

```bash
➜ mkdir lhci-test && cd lhci-test
➜ yarn init -y && yarn add @lhci/cli
➜ yarn lhci collect -n=1 --url=https://example.com
Running Lighthouse 1 time(s) on https://example.com
Run #1...done.
Done running Lighthouse!
➜ cat > lighthouserc.js <<EOF
module.exports = {ci: {assert: {assertions: {'works-offline': 'error'}}}}
EOF
➜ yarn lhci assert
Checking assertions against 1 URL(s), 1 total run(s)

1 result(s) for https://example.com/ :

  ✘  works-offline failure for minScore assertion
       Current page does not respond with a 200 when offline
       https://web.dev/works-offline

        expected: >=1
           found: 0
      all values: 0

Assertion failed. Exiting with status code 1.
error Command failed with exit code 1.
➜ cat > lighthouserc.js <<EOF
module.exports = {assert: {assertions: {'works-offline': 'error'}}}
EOF
➜  yarn lhci assert
Error: No assertions to use
    at Object.runCommand (/Users/patrick/Code/Playgrounds/foo/node_modules/@lhci/cli/src/assert/assert.js:51:48)
    at run (/Users/patrick/Code/Playgrounds/foo/node_modules/@lhci/cli/src/cli.js:88:23)
    at Object.<anonymous> (/Users/patrick/Code/Playgrounds/foo/node_modules/@lhci/cli/src/cli.js:119:1)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:754:12)
error Command failed with exit code 1.
```

The code you linked is for processing the options after it has been through the yargs CLI config processing which flattens all the `ci` properties.

https://github.com/GoogleChrome/lighthouse-ci/blob/ab104a9221de95efe5c7d31f473efc41ec157211/packages/utils/src/lighthouserc.js#L153-L160

skippednote:
@patrickhulce Just realised that my assert was nested under collect, not ci. 🤦

PR #326, opened 2020-05-26
feat(server): support advanced sequelize options

cannjeff:
Adds the ability to provide advanced configuration options for Sequelize.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1043,6 +1043,30 @@ module.exports = {
 };
 ```
 
+### Custom Sequelize Configuration
+
+See [the sequelize docs](https://sequelize.org/v4/) for more information on available settings.
+
+```js
+module.exports = {
+  ci: {
+    server: {
+      storage: {
+        storageMethod: 'sql',
+        sqlDialect: 'postgres',
+        sqlConnectionSsl: true,
+        sqlConnectionUrl: process.env.DATABASE_URL,
+        sequelizeOptions: {
+          pool: {
+            acquire: 30000,
+          },
+        },
+      },
+    },
+  },
+};
+```
+
 ### Custom Headers for Wizard
 
 If you're running the `lhci server` behind a reverse proxy or any other component that requires some extra headers you can configure them in the wizard section `extraHeaders`.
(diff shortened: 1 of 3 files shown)


packages/server/src/api/storage/sql/sql.js
```diff
@@ -80,6 +80,7 @@ function createSequelize(options) {
   const sequelizeOptions = {
     operatorsAliases: false,
     logging: () => {},
+    ...options.sequelizeOptions,
```
patrickhulce, line 83: let's at least move `operatorsAliases: false` down below this so it can't be overriden, we don't want folks to enable that and all of the work has already been done in LHCI to make due without them

docs/configuration.md
```diff
@@ -1043,6 +1043,30 @@ module.exports = {
 };
 ```
 
+### Custom Sequelize Configuration
+
```
patrickhulce, line 1047:
```suggestion

See [the sequelize docs](https://sequelize.org/v4/) for more information on available settings.

```

docs/configuration.md
```diff
@@ -1043,6 +1043,30 @@ module.exports = {
...
+```js
+const fs = require('fs');
+
```
patrickhulce, line 1050:
```suggestion
```

this example doesn't use `fs`

packages/server/src/api/storage/sql/sql.js
cannjeff: Good call. I wasn't sure which settings you specifically wouldn't want configurable. What about `logging` does it make sense to leave that overridable?

docs/configuration.md
cannjeff: That's what I get for copy/paste... Good catch.

packages/server/src/api/storage/sql/sql.js
patrickhulce: `logging` is fine with me, it's just unnecessary noise but if folks want more insight into what it's doing that's fine with me :)

cannjeff:
Address all comments. Thanks!

patrickhulce:
thanks @cannjeff! 🎉

PR #328, opened 2020-05-27
feat(cli): add filesystem upload target

patrickhulce:
closes #142 

Basic usage: `lhci upload --target=filesystem --outputDir=./the-reports`

```
|- (cwd)
  |- the-reports
    |- manifest.json
    |- www_example_com-_page-2020_05_22_21_15_05.report.json
    |- www_example_com-_page-2020_05_22_21_15_05.report.html
```

```jsonc
[
  {
    "url": "https://www.example.com/page",
    "isRepresentativeRun": true, // whether it's the median run
    "htmlPath": "/path/to/cwd/the-reports/www_example_com-_page-2020_05_22_21_15_05.report.html",
    "jsonPath": "/path/to/cwd/the-reports/www_example_com-_page-2020_05_22_21_15_05.report.json",
    "summary": {"performance": 0.95, "seo": 0.9, "best-practices": 1}
  }
]
```

Issue #142, opened 2019-12-06
Upload target filesystem, Override File Naming


alexfornuto:
Thanks to #139 I can now handle the JSON and HTML reports locally without having to upload the results to an LHCI server (thanks for that!).

In order to programmatically handle the files generated, I'd like to be able to define a consistent pattern for the files. For example:

IF I run `lchi collect --url=https://pantheon.io/docs --url=https://some.staging.environment.io/docs`,
AND I define some extra parameter like `--format=$URL-$PASS_NUM` (just spit-balling on this)
THEN the contents of `.lighthouseci` would look like:

```
.lighthouseci
|- lhr-https://pantheon.io/docs-1.html
|- lhr-https://pantheon.io/docs-1.json
|- lhr-https://pantheon.io/docs-2.html
|- lhr-https://pantheon.io/docs-2.json
|- lhr-https://pantheon.io/docs-3.html
|- lhr-https://pantheon.io/docs-3.json
|- lhr-https://some.staging.environment.io/docs-1.html
|- lhr-https://some.staging.environment.io/docs-1.json
|- lhr-https://some.staging.environment.io/docs-2.html
|- lhr-https://some.staging.environment.io/docs-2.json
|- lhr-https://some.staging.environment.io/docs-3.html
|- lhr-https://some.staging.environment.io/docs-3.json
```

Not that the filename _needs_ to contain the full URL, this is just an example where it uses a parameter already defined in another flag. Anything consistently predictable would suite my needs.

jzabala:
Hey @alexfornuto is the incremental number sequence on the files something that you are looking for too?

In your example I see you used a pattern: {url}-{sequential-number}, but a result from the lib looks more like this:

![Screen Shot 2019-12-06 at 1 44 10 PM](https://user-images.githubusercontent.com/1315054/70343915-d176af80-182e-11ea-8994-f1710db70085.png)

Or based on the image the only thing you want to change is the prefix `lhr-`?

alexfornuto:
Sorry for not being more precise. The numbers in my example correlate to the three passes I've configured on my end. Since I know that there are three passes, those numbers are predictable, whereas the current timestamp format is not.

I don't care about the `lhr` prefix, and my example includes it. The main point here is that the timestamp data can't be predicted, and I'd like to override it.

patrickhulce:
[...]

PR #330, opened 2020-05-28
fix(cli): throw on existing manifest.json

patrickhulce:
followup to #328 and #142 

prevents accidental overwriting and misconceptions about merging manifest.json files when using target=filesystem


patrickhulce:
This option lost in the poll compared to the prior behavior and no one has complained about it getting overwritten, so I guess we'll just stick with it for now 🤷

PR #337, opened 2020-06-01
Make build context related errors more verbose.

szwarckonrad:
The issue I faced during not so standard implementation of lhci was error thrown by getAvatarUrl function. It stated that the app wasn't able to `determine commit email with ...`. and it made me look at LHCI_BUILD_CONTEXT__AUTHOR because that was the only place any email was mentioned. Only by looking at source code I saw that the env I was actually looking for was LHCI_BUILD_CONTEXT__AVATAR_URL, hence the better explanation in error's message.  Additionally, for every error that can be hardcoded with setting envs I've specified what env can be used to override it.

### DIFF
--- a/packages/utils/src/build-context.js
+++ b/packages/utils/src/build-context.js
@@ -71,7 +71,10 @@ function getCurrentHash() {
     encoding: 'utf8',
   });
   if (result.status !== 0) {
-    throw new Error('Unable to determine current hash with `git rev-parse HEAD`');
+    throw new Error(
+      'Unable to determine current hash with `git rev-parse HEAD`. ' +
+        'This can be overridden with setting LHCI_BUILD_CONTEXT__CURRENT_HASH env.'
+    );
   }
 
   return result.stdout.trim();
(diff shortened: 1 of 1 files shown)


packages/utils/src/build-context.js
```diff
@@ -216,7 +231,10 @@ function getAvatarUrl(hash = 'HEAD') {
     encoding: 'utf8',
   });
   if (result.status !== 0) {
-    throw new Error('Unable to determine commit email with `git log --format=%aE -n 1`');
+    throw new Error(
+      "Couldn't get commit author's avatar url because application was unable to determine commit email with `git log --format=%aE -n 1`. " +
```
patrickhulce, line 235:
```suggestion
      "Unable to determine commit author's avatar URL because `git log --format=%aE -n 1` failed to provide author's email. " +
```

feels slightly more inline with the way the others are worded, but not a big deal

patrickhulce: The extra push to fix should also retry the tests which look like a flaky failure of puppeteer :)

patrickhulce: This was marked as resolved but no explanation. Do you prefer to keep the existing wording? If so I'm curious why :)

szwarckonrad: > This was marked as resolved but no explanation. Do you prefer to keep the existing wording? If so I'm curious why :)

Sorry, forgot to click `Commit suggestion`. Of course if the wording you proposed better suits the pattern of the app then sure :)

PR #342, opened 2020-06-03
feat(server): add support for autocollect PSI results

### LINKED ISSUE #5: Support production monitoring use case, no git required
Currently every build needs a `.hash` but we don't necessarily need one. We could switch to a `.projectVersion` and support timestamps instead?

patrickhulce:
Opening for discussion on the API for production monitoring of URLs

~~After starting to implement, I'm thinking something like this. I know the code doesn't match right now, that's why I'm opening this draft :)~~

```js
module.exports = {
  ci: {
    server: {
      psiCollectCron: {
        psiApiKey: process.env.PSI_API_KEY,
        sites: [
        {
          label: 'Example',
          buildToken: 'firstbuildtoken',
          schedule: '* * * * 5', // cron syntax
          numberOfRuns: 5,
          urls: ['http://example.com', 'http://example.com/subpage'],
        },
        {
          label: 'Other Site',
          buildToken: 'secondbuildtoken',
          schedule: '* * * * 5', // cron syntax
          numberOfRuns: 3,
          urls: ['http://othersite.com'],
        },
        ]
      },
    }
  }
}
```

fixes #5


connorjclark:
sounds cool.

needs a `name` field too.

should think about how thresholds might be defined too, for showing on the graphs / potential future alerting feature.

patrickhulce:
> needs a name field too.

What would a name field do?

> should think about how thresholds might be defined too, for showing on the graphs / potential future alerting feature.

I think this is out of scope for now :)

patrickhulce:
To clarify, you would need to create a project already for this to work, that's where the `buildToken` comes from.

connorjclark:
> What would a name field do?

You could have a thousands of URLs for a page type with different content. There may be a way to group them by features that are enabled on a subset. You'd want to pick a few from each, add to a monitoring group, and give a name. ex: "Articles with Interactive Graphs"; "Normal Content Articles"; "Articles behind paywall"

patrickhulce:
> You'd want to pick a few from each, add to a monitoring group, and give a name

For this to work in the current LHCI environment, there will be at most 1 entry in `monitor.projects` per project. Given that a project already has a name, I'm wondering what value a monitoring group name would provide that isn't served by adding a comment in the config. 

My concern is that it would cause confusion that it creates a project, or looks up the project by name, etc.

patrickhulce:
Also very open to ideas on how change the config to make that relationship more obvious if that's where the miscommunication was :)

paulirish:
IMO it's separate from this API, but I'm also attracted to user-defined names for these URLs.  would be much easier to see "Homepage (mobile)" and "Homepage (desktop)" than what you currently gotta look at. Perhaps the server config accepts a UrlToName hash.

paulirish:
it does seem slightly odd that prod monitoring isn't part of `collect` configuration.

what about a new toplevel config section?

alekseykulikov:
Great idea!

A few notes:

- `url` could be a string or an object with an optional `name` property, ex:
```js
urls: [{ name: "Homepage", url: 'https://example.com' }, 'https://example.com/blog']
```
- it's already possible to monitor with a cron job or something like [Github Actions `schedule`](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule). Is this functionality is more about adding `cron` to the server? Or is it a new section with a possibility of alerting and other monitoring features? The difference between `monitor` and `collect` is not clear.

patrickhulce:
Ooooooh a name *per URL* not *per group*. Sorry for the misunderstanding @connorjclark. 

Yeah that sounds great generally though I agree completely separate from this API. We can/should explore that regardless of performance monitoring.

> it does seem slightly odd that prod monitoring isn't part of collect configuration.
what about a new toplevel config section?

It will only function as part of the server though. The idea is your LHCI server *is* your collection agent using PSI. Do you think it would be confusing to have a separate toplevel section that only works when using the `server` command? It would kinda break the 1:1 mapping of configuration to command we have at the moment. 

> Is this functionality is more about adding cron to the server?

Correct, it's about getting regular results for your URLs with the LHCI server being your *only* infrastructure. You can definitely accomplish everything this feature would give you today by standing up a separate server with a cron.

There's definitely more that *could* be done with alerts and such, but that's beyond the scope of what we're accomplishing right now.

connorjclark:
> Ooooooh a name per URL not per group. Sorry for the misunderstanding @connorjclark.

No, I was suggesting per group. why not both.

patrickhulce:
Alright digesting the feedback, I think the discussion headed a direction I wasn't tackling because of the "monitor" language used, for now I'll rename this property to `autocollect` to better reflect the fact that it's not true monitoring and it's just having the server do some features of collect for you. (For example: alerts should be able to be configured completely independently of having the server use PSI to automatically collect results, you might want to be alerted for regular CI threshold failures, etc)

The rest of the suggestions for future work should be captured by other issues now 👍  thanks everyone!

patrickhulce:
Talked with @paulirish and we're gonna rename this key super specific `psiCronCollect`, leave room for `autocollect` as a separate toplevel property+command

patrickhulce:
alright I think all feedback has been addressed or capture in separate issues, I intend to merge this tomorrow if there are no major objections

yuriteixeira:
Nice! Looking forward to get a release including this :)

patrickhulce:
it's been released @yuriteixeira :) 0.4.2

Issue #352, opened 2020-06-18
Unable to upload project by build token


jflayhart:
After running `lhci wizard` successfully, I get the resulting build token to use for uploading. Collecting data with `lhci autorun` works as expected except the get project by token API doesn't work (404):

```
// client cli config
upload: {
      target: 'lhci',
      serverBaseUrl: 'https://custom-server.com/lighthouse',
      token: TOKEN,
      basicAuth: {
        username,
        password
      }
    }

✅  .lighthouseci/ directory writable
✅  Configuration file found
✅  Chrome installation found
✅  Ancestor hash determinable
✅  LHCI server reachable
⚠️   LHCI server not API-compatible
⚠️   LHCI server token invalid
✅  LHCI server unique build for this hash
Healthcheck passed!

....successful audit completes

Error: Could not find active project with provided token
    at runLHCITarget (/Users/jflayhart/.nvm/versions/node/v13.12.0/lib/node_modules/@lhci/cli/src/upload/upload.js:380:11)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Object.runCommand (/Users/jflayhart/.nvm/versions/node/v13.12.0/lib/node_modules/@lhci/cli/src/upload/upload.js:572:16)
    at async run (/Users/jflayhart/.nvm/versions/node/v13.12.0/lib/node_modules/@lhci/cli/src/cli.js:92:7)
WARNING: upload command failed.
assert command failed. Exiting with status code 1.
```

I have curl'd some project endpoints y'all use in your source code and /projects returns this:

```
// GET https://custom-server.com/lighthouse/v1/projects
[
    {
        "id": "123-abc",
        "name": "repo-name",
        "token": "", // empty!
        "adminToken": "", // empty (although not set in GUI)
        ...
    }
]
```

Any ideas what could be going on with project creation? I am definitely using the correct build token that was provided from wizard and PostGres is stable.

**EDIT:** I just found you clear the tokens in response intentionally, which is sane: https://github.com/GoogleChrome/lighthouse-ci/blob/master/packages/server/src/api/routes/projects.js#L27

So I tried `POST https://custom-server.com/lighthouse/v1/projects/lookup` with token and get 200:

```
{
    "id": "123-abc",
    "name": "repo-name",
    "token": "ACTUAL TOKEN",
    ...
}
```

lol what gives?

jflayhart:
[...]

Issue #308, opened 2020-05-14
Document that LHCI server must own the entire origin


niklasgrahl:
Hi, I'm trying to run the lhci server on a url like 
```
http://dev-server.example.org/dev-tools/lighthouse-ci-server/app/
```
but the server tries to load everything relative to 
```
http://dev-server.example.org/app/
```
 for example `http://dev-server.example.org/app/entry.fd7e7398.js`. It also tries to redirect from `http://dev-server.example.org/dev-tools/lighthouse-ci-server/` to `http://dev-server.example.org/app/`. I can't find any way of setting the asset path at the moment but maybe I am missing something.

**Would you be open to adding a base/public url or asset prefix config option for the server to support this use case?** 

I guess the mount path could be read on server startup but it looks like the asset path is [set at build time](https://github.com/GoogleChrome/lighthouse-ci/blob/b595264ac67dec5afc8cc0b5eaac295ae50f8ac4/packages/server/package.json#L16) so I'm not sure how easy this would be to configure at startup.

It looks like the upload command might already support this use case with the `serverBaseUrl` option, but I haven't tested it. It would be great to have it for the server as well.

Thanks!

patrickhulce:
Thanks for filing @niklasgrahl!

This is a thorny problem that would touch many layers of Lighthouse CI. Not only is the UI affected by every API client as well. The existing `serverBaseUrl` option you point out effectively only supports setting the origin at the moment. In fact I wonder if we should rename this to `serverOrigin` in 0.4.x to make that clearer 🤔 

You're also correct that because the base path is set at build time and we have SPA fake URL paths, any changing of the base path would require a custom build of the frontend code (or a significant rework of how the app references static assets).

Given all of these constraints and the considerable increase in testing surface it would require, I'm not sure this is something we can realistically consider for the near future of the project :/

Happy to add whatever you think would be clear to the server documentation to make this limitation more upfront and obvious!

niklasgrahl:
Thanks for the quick reply! That's ok, I'll see if I can deploy it on a separate subdomain instead 👍 

Cheers!

nick-rudenko:
@niklasgrahl have you managed to make it work using a subdomain? Having the same issue

patrickhulce:
@nick-rudenko subdomain shouldn't require anything special, all the LHCI servers I know of use a subdomain. What issue are you running into with subdomain?

nick-rudenko:
[...]

PR #353, opened 2020-06-18
[RFC] Fix base url issue where the wrong origin is used

jflayhart:
Fixes: https://github.com/GoogleChrome/lighthouse-ci/issues/352
Fixes: #308

Based on the following lhrc config:

```
{
  ci: {
    upload: {
      target: 'lhci',
      serverBaseUrl: 'https://custom-server.com/lighthouse',
      token: TOKEN,
      basicAuth: {
        username,
        password
      }
    }
  }
}
```

## Problem
`_normalizeUrl` resolves to origin url set by `serverBaseUrl`, not the desired `serverBaseUrl` as set:
`https://custom-server.com`

## Solution
I propose that the `serverBaseUrl` for client CLI config is always used when the target is `lhci`. If someone wants `domain.com/path/to/server` then it should honor that, not use `domain.com`

Always resolve relative to `serverBaseUrl` not the origin:
`https://custom-server.com/lighthouse`

### DIFF
--- a/docs/recipes/lhci-server-vpn-proxy/README.md
+++ b/docs/recipes/lhci-server-vpn-proxy/README.md
@@ -1,0 +1,47 @@
+
+# nginx-based LHCI Proxy Server
+
+**NOTE: be sure to read the [Security section](../../server.md#Security) of the server documentation to protect your server properly**
+
+## Overview
+
+This assumes some knowledge of Docker/K8 (see [docker recipe](../docker-server)), VPN remote access, and [nginx config](https://www.nginx.com/resources/wiki/start/topics/examples/full/). It's not uncommon that a company develops internal tools behind VPN, but if you want to intentionally expose an internal service, like LHCI, you typically use [a proxy pattern](https://en.wikipedia.org/wiki/Proxy_pattern).
+
+This is a contrived recipe on how to configure a nginx reverse proxy to expose parts of your LHCI server that exists behind a VPN.
+
+## Architecture
+
+Say you build a custom LHCI server behind your company's VPN:
+```
+|-----VPN-----|
+| LHCI server | X-- www <-- client ¯\_(⊙︿⊙)_/¯
+|             |
+|-------------|
+```
+You hand out VPN profiles to employees to access this private LHCI server at `https://lhc-over-vpn.example.com` and you see the LHCI server GUI in all its glory:
+```
+|-----VPN-----|
+| LHCI server | <-- VPN tunnel <-- www <-- remote access client ヽ(´ー｀)ノ
+|             |
+|-------------|
+```
+### Problem
+
+That works when viewing the GUI with a browser at `https://lhc-over-vpn.example.com`, but what happens when you need to provide access for public-facing CI tools to upload reports to your private LHCI server? For example, CI tools like [CircleCI](https://circleci.com/blog/vpns-and-why-they-don-t-work/) and [TravisCI](https://docs.travis-ci.com/user/common-build-problems/#ftpsmtpother-protocol-do-not-work) have limitations in dealing with VPN tunneling or else you are stuck with extremely complicated docker setup using `ssh` and `openvpn`. It's definitely not supported out of the box and doesn't work at all with VPN + MFA.
+
+### A Proxy Solution
+
+So you use a CI tool like CircleCI and want to allow uploading reports to your private LHCI server. One solution might be to use a nginx reverse proxy server to [expose only certain routes](https://github.com/GoogleChrome/lighthouse-ci/blob/v0.4.1/packages/server/src/server.js#L51) for your CI tool to communicate with your LHCI server.
+
+⚠️ It is recommended you at least password-protect using [LHCI basic auth](../../server.md#basic-authentication), since you will be exposing these routes to the public.
+
+So at the gateway of your company's VPN, you might have a nginx server where you can reverse proxy to your private LHCI server:
+```
+|-----VPN-----|
+| LHCI server | <-- VPN tunnel <-- lhc-over-vpn.example.com/app/projects <-- remote access client
+|             |
+|             | <-- upload <-- (lhc-over-vpn.example.com/v1/projects) <-- www.public-proxy.example.com/lighthouse/v1/projects <-- CircleCI
+|-------------|
+```
+
+With this setup, the GUI at `lhc-over-vpn.example.com/app/projects` is left intact behind the VPN, but external CI tools now have public access to the exposed, but still password-protected, reverse proxy server (`www.public-proxy.example.com/lighthouse/v1`)!
--- a/docs/recipes/lhci-server-vpn-proxy/nginx.conf
+++ b/docs/recipes/lhci-server-vpn-proxy/nginx.conf
@@ -1,0 +1,18 @@
+server {
+    listen      443;
+    server_name public-proxy.example.com;
+    # required to pass LHCI CLI healthcheck
+    location = /lighthouse/version {
+        proxy_set_header Host  https://lhc-over-vpn.example.com;
+        proxy_ssl_name         https://lhc-over-vpn.example.com;
+        set $upstream_endpoint https://lhc-over-vpn.example.com;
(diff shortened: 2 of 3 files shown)


patrickhulce:
Thanks very much for the PR! :)

@jflayhart mind explaining how you got the server UI to work while hosting the server on a subpath? The public asset and link references are all hardcoded to `/app/...`.

I read https://github.com/GoogleChrome/lighthouse-ci/issues/308#issuecomment-646205522 but I didn't understand the proxy explanation 😊 

I'm concerned about supporting this if the UI will then still be broken. The current non-support was intentional for this reason.

jflayhart:
tenor

haha seriously though, we operate behind a VPN, so the server and its app GUI are running on a separate container behind that VPN: `https://lighthouse-server.com`

Naturally, you must be on VPN to access `https://lighthouse-server.com`, so when you are that works just fine via that URL. But we can't expect other contributors to always be on our VPN and arguably we shouldn't need to make a VPN profile for each contributor to upload, so I publicly exposed only `https://lighthouse-server.com/version` and `https://lighthouse-server.com/v1/*` behind basic auth instead, respectively:

- `https://custom-server.com/lighthouse/version`
- `https://custom-server.com/lighthouse/v1/*`

So enter client-side config... I can only upload to and accass those two routes with basic auth. I get this working with curl, but then this is where the client config `serverBaseUrl` breaks down.

jflayhart:
Maybe you think I'm conflating things with client and server, but I am saying that the server is decoupled from any client config.

- some client has its own CLI lighthouserc.js config
- single lighthouse ci server has its own dedicated lighthouserc.js config that doesn't care about any of this proxy stuff.

patrickhulce:
Gotcha, very cool! If I could interest you in a deal to document the proxy UI bit for how to set something like that up, I'd be happy to fix up the server upload piece of things you've already started here :)

jflayhart:
Sure. Honestly running this on a subdomain would be easier or ideally everyone can connect over VPN. Sadly, life doesn't always work out the way we want it to ;)

How would you want me to contribute to the docs? Like a disagram? I am using nginx and stuff, but software like yours is meant to be generic and service agnostic.

patrickhulce:
> I am using specific services that may not apply to everyone, but software like yours is meant to be generic and service agnostic.

That's OK! I'm just thinking a note in our server documentation `WARNING: things will be much easier for you if you let LHCI own the entire origin, if you can't or don't want to go read this example of how to configure things in our recipes.` and then link off to a doc you write in `docs/recipes/lhci-server-vpn-proxy` with a little diagram and quick explanation of however you accomplished it. No need to make it ultra generic or anything, recipes is about showing examples of how to use LHCI in specific real-world scenarios.

docs/recipes/lhci-server-vpn-proxy/README.md
```diff
@@ -0,0 +1,47 @@
...
+|             |
+|-------------|
+```
+### Problem
+
+That works when viewing the GUI with a browser at `https://vpn.lhci.com`, but what happens when you need to provide access for public CI tools to upload reports to your private LHCI server? For example, CI tools like [CircleCI](https://circleci.com/blog/vpns-and-why-they-don-t-work/) and [TravisCI](https://docs.travis-ci.com/user/common-build-problems/#ftpsmtpother-protocol-do-not-work) have limitations in dealing with VPN tunneling or else you are stuck with extremely complicated docker setup using `ssh` and `openvpn`. It's definitely not supported out of the box if at all.
+
+### A Proxy Solution
+
+So you use a CI tool like CircleCI and want to allow uploading reports to your private LHCI server. One solution might be to use a nginx reverse proxy server to [expose only certain routes](https://github.com/GoogleChrome/lighthouse-ci/blob/master/packages/server/src/server.js#L51) for your CI tool to communicate with your LHCI server.
```
patrickhulce, line 34: I don't want this line to get out of date, how about a specific tag or ref?
```suggestion
So you use a CI tool like CircleCI and want to allow uploading reports to your private LHCI server. One solution might be to use a nginx reverse proxy server to [expose only certain routes](https://github.com/GoogleChrome/lighthouse-ci/blob/<pick your favorite ref or tag>/packages/server/src/server.js#L51) for your CI tool to communicate with your LHCI server.
```

docs/recipes/lhci-server-vpn-proxy/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+⚠️ It is recommended you at least password-protect using [LHCI built auth](https://github.com/jflayhart/lighthouse-ci/blob/patch-1/docs/server.md#basic-authentication), since you will be exposing these routes to the public.
```
patrickhulce, line 36:
```suggestion
⚠️ It is recommended you at least password-protect using [LHCI basic auth](../../server.md#basic-authentication), since you will be exposing these routes to the public.
```

docs/recipes/lhci-server-vpn-proxy/README.md
```diff
@@ -0,0 +1,47 @@
...
+So at the gateway of your company's VPN, you might have a nginx server where you can reverse proxy your private LHCI server to be exposed over your public gateway:
+```
+|-----VPN-----|
+| LHCI server | <-- VPN tunnel <-- www <-- remote access client
+|             |
+|             | <-- upload <-- vpn.lhci.com/v1/projects <-- www.api-proxy.com/lighthouse/v1/projects <-- CircleCI
+|-------------|
+```
+
+With this setup the GUI (vpn.lhci.com/app/projects) is left in tact behind the VPN, but things like CI tools now have public access to the exposed (password-protected), reverse proxy server (www.api-proxy.com/lighthouse/v1)!
```
patrickhulce, line 47:
```suggestion
With this setup, the GUI (vpn.lhci.com/app/projects) is left intact behind the VPN, but external CI tools now have public access to the exposed, but still password-protected, reverse proxy server (www.api-proxy.com/lighthouse/v1)!
```

docs/recipes/lhci-server-vpn-proxy/nginx.conf
```diff
@@ -0,0 +1,18 @@
+server {
+      listen      443;
+      server_name api-proxy.com;
```
patrickhulce, line 3:
```suggestion
      server_name public-proxy.example.com;
```

we try to stick to <subdomain>.example.com to ever prevent accidentally sending someone somewhere nefarious :)

maybe search and replace `api-proxy.com` with `public-proxy.example.com` and `vpn.lhci.com` with `lhc-over-vpn.example.com`?

docs/server.md
```diff
@@ -21,6 +21,8 @@ The LHCI server can be run in any node environment with persistent disk storage
 
 If you'd like to run LHCI on your own infrastructure, you'll need a package.json to install the dependencies via npm. See the [Heroku](./recipes/heroku-server/README.md) and [docker](./recipes/docker-server/README.md) recipes for more examples on how to deploy the LHCI server.
 
+:warning: Things will be much easier if you let LHCI own the entire origin of your domain, however, if you can't do that and have to operate over a proxy [read this example of how to configure a proxy in our recipes](./recipes/lhci-server-vpn-proxy/README.md).
```
patrickhulce, line 24:
```suggestion
**NOTE:** LHCI really needs to own the entire origin to work as expected. However, if you can't do that and have to operate over a proxy, you can follow the [example of how to configure LHCI on a subpath](./recipes/lhci-server-vpn-proxy/README.md).
```

I know I wrote that originally but feels a bit odd as a warning at this point, more like a note :)

packages/utils/src/api-client.js
```diff
@@ -46,6 +46,7 @@ class ApiClient {
    * @return {URL}
    */
   _normalizeURL(url) {
+    if (this._rootURL) return new this._URL(this._rootURL + url, this._rootURL);
```
patrickhulce, line 49: you can go ahead and revert this for the tests I'll make a pass and clean this up with some tests too :)

jflayhart: reverted and squashed my changes

PR #367, opened 2020-07-02
Clarify puppeteer docs

### LINKED ISSUE #361: Document how an array of urls interacts with the puppeteerScript flag
Hi!

I've been trying to set up a puppeteer-script to bypass authentication. However, I am having some trouble figuring out how the puppeteerScript flag interacts with the URL list I have.
**lighthouserc.json**

```
      "puppeteerScript": "puppeteer-login-script.js",
      "url": [
        "http://127.0.0.1:8181/",
        "http://127.0.0.1:8181/about",
        "http://127.0.0.1:8181/contact",
        "http://127.0.0.1:8181/donate",
        "http://127.0.0.1:8181/get-started",
        "http://127.0.0.1:8181/community-library",
        "http://127.0.0.1:8181/privacy-policy",
        "http://127.0.0.1:8181/teach",
        "http://127.0.0.1:8181/terms",
        "http://127.0.0.1:8181/thanks"
      ]
    },
```

I can get it to work with one URL, but I think my config currently runs the Puppeteer script before each URL collection and keeps trying to log in even though I am already logged in. Is there a way I make it so that the `lighthouserc` config will only run the script on certain URLs. Or do I need to handle it in the `puppeteer-login-script`

jflayhart:
Fixes #361

I still think there's more nuance to this script stuff, but hopefully this covers in more detail?

Preview: https://github.com/GoogleChrome/lighthouse-ci/pull/367/files?short_path=76e7313#diff-76e731333fb756df3bff5ddb3b731c46

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -305,9 +305,6 @@ module.exports = async (browser, context) => {
   await page.type('#password', 'password');
   await page.click('[type="submit"]');
   await page.waitForNavigation();
-  // we're logged in, so test the page behind auth
-  // runs the url set in lighthouserc
-  await page.goto(context.url);
   // close session for next run
   await page.close()
 };
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -274,7 +274,21 @@ An optional path to a JavaScript file that exports a function that uses puppetee
 
 **NOTE:** In order to use puppeteer scripts, you need to install `puppeteer` yourself, i.e. `npm install --save puppeteer`. Any version you like that has a `.launch` method compatible with v1.x or v2.x and works with the Chrome version installed in your environment should work.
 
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -284,16 +298,22 @@ When running Lighthouse CI on a page with behind authentication, you'll need to
...
+  // launch browser for LHCI
   const page = await browser.newPage();
   await page.goto('http://localhost:8080/login');
   await page.type('#username', 'admin');
   await page.type('#password', 'password');
   await page.click('[type="submit"]');
   await page.waitForNavigation();
+  // we're logged in, so test the page behind auth
+  // runs the url set in lighthouserc
+  await page.goto(context.url);
```
patrickhulce, line 310: this shouldn't be necessary because Lighthouse is going to open a new tab and navigate to this page immediately after this script finishes.

was there something you ran into that prompted this?

patrickhulce, line 310:
```suggestion
```

jflayhart: yeh, I could be very wrong though. Somehow I got to a point where this was necessary because the page was empty and the LHCI run would simply time out. I threw this in there and it worked.

Regardless, I agree this probably isn't necessary as a generic example!

Issue #385, opened 2020-07-21
Adding support for socket path


MaximeHeckel:
Hey 👋 

My team and I were trying to spin up lighthouse server on Google Cloud Run instead of GKE (we think that setting up GKE might be a bit overkill for our needs) and wanted to use a Postgres instance on Google Cloud SQL to save our report data.
However since Cloud Run instances are ephemeral, we can't rely on the `sqlConnectionURL` to connect to Google Cloud SQL (you have to manually whitelist IPs that can initiate a connection to the DB via public IP). 
We usually rely on socket paths for our DB connections and were wondering if we could start a PR on this project to allow the `lighthouserc.json` file to support some socket path options 😄 (we have a PoC working as I'm writing these words)

patrickhulce:
Sure @MaximeHeckel sounds good, happy to accept a PR! I think testing might be difficult, but we can hopefully figure something out.

PR #389, opened 2020-07-24
feat(server): support socket path in server config

MaximeHeckel:
Fixes https://github.com/GoogleChrome/lighthouse-ci/issues/385

- [x] Add `sqlSocketPath` option support in the config
- [x] Validate config UX with project maintainers
- [x] Add tests for new socket path option in config (Looking for guidance here 😄 )
- [x] Mention `sqlSocketPath` in `docs/configuration.md`

As mentioned in the ticket, my team and I were looking to set up lighthouse-ci server on Google Cloud Run (serverless containers) with a Google Cloud SQL database (Postgres) and our only possibility to link the two was to use a socket path instead of a connection URL.

Let me know what you think of this config, there's probably a few things that still need to be done or validated in terms of how you want the config to look like.
Also @patrickhulce mentioned that testing might be quite hard for this one, so I'm looking for any proposal on how I could tackle this 😄

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1205,6 +1205,26 @@ module.exports = {
 }
 ```
 
+### Socket Path for Database Connection
+
+```json
+{
+  "ci": {
+    "server": {
+      "storage": {
+        "sqlDialect": "mysql",
+        "sqlSocketPath": "/var/lib/mysql/mysql.sock",
+        "sequelizeOptions": {
+          "database": "reports",
+          "username": "admin",
+          "password": "password"
+        }
+      }
+    }
+  }
+}
+```
+
 ### Custom SSL Certificate for Database Connection
 
 ```js
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -892,15 +892,6 @@ _sqlDialect=sqlite only_
 
 The path to the sqlite database on the local filesystem relative to the current working directory.
 
-##### `storage.sqlSocketPath`
-
-_sqlDialect=mysql or sqlDialect=postgres only_
-
-The database Unix socket file path for the MySQL or PostgreSQL database.
-
-Note: If used, this option will require to specify the `database`, `username` and `password` sequelize options.
-See `storage.sequelizeOptions` for more information.
-
 ##### `storage.sqlConnectionUrl`
 
 _sqlDialect=mysql or sqlDialect=postgres only_
@@ -1213,7 +1204,9 @@ module.exports = {
     "server": {
       "storage": {
         "sqlDialect": "mysql",
-        "sqlSocketPath": "/var/lib/mysql/mysql.sock",
+        "sqlDialectOptions": {
+          "socketPath": "/var/lib/mysql/mysql.sock"
+        },
         "sequelizeOptions": {
           "database": "reports",
           "username": "admin",
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -306,7 +306,7 @@ module.exports = async (browser, context) => {
   await page.click('[type="submit"]');
   await page.waitForNavigation();
   // close session for next run
-  await page.close()
+  await page.close();
 };
 ```
 
@@ -891,6 +891,15 @@ One of `mysql`, `postgres`, or `sqlite`. `sqlite` in a local file on disk has be
 _sqlDialect=sqlite only_
 
 The path to the sqlite database on the local filesystem relative to the current working directory.
+
+##### `storage.sqlSocketPath`
+
+_sqlDialect=mysql or sqlDialect=postgres only_
+
+The database Unix socket file path for the MySQL or PostgreSQL database.
+
+Note: If used, this option will require to specify the `database`, `username` and `password` sequelize options.
+See `storage.sequelizeOptions` for more information.
 
 ##### `storage.sqlConnectionUrl`
 
--- a/packages/server/src/api/storage/sql/sql.js
+++ b/packages/server/src/api/storage/sql/sql.js
@@ -92,6 +92,32 @@ function createSequelize(options) {
     });
   }
 
+  if (options.sqlSocketPath) {
+    if (
+      !options.sequelizeOptions ||
+      !options.sequelizeOptions.database ||
+      !options.sequelizeOptions.username ||
+      !options.sequelizeOptions.password
+    ) {
+      throw new Error(
+        `Cannot use ${dialect} with socketPath without database, username, or password`
+      );
+    }
+
+    return new Sequelize(
+      options.sequelizeOptions.database,
+      options.sequelizeOptions.username,
+      options.sequelizeOptions.password,
+      {
+        ...sequelizeOptions,
+        dialect: options.sqlDialect,
+        host: options.sqlSocketPath,
+        ssl: !!options.sqlConnectionSsl,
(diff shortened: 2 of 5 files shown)


MaximeHeckel:
@googlebot I signed it!

docs/configuration.md
```diff
@@ -892,6 +892,15 @@ _sqlDialect=sqlite only_
 
 The path to the sqlite database on the local filesystem relative to the current working directory.
 
+##### `storage.sqlSocketPath`
+
+_sqlDialect=mysql or sqlDialect=postgres only_
+
+The database Unix socket file path for the MySQL or PostgreSQL database.
+
+Note: If used, this option will require to specify the `database`, `username` and `password` sequelize options.
```
patrickhulce, line 901: A complete example configuration in the examples section would be nice to link to :)

packages/server/src/api/storage/sql/sql.js
```diff
@@ -92,6 +92,31 @@ function createSequelize(options) {
...
+  if (options.sqlSocketPath) {
+    if (
+      !options.sequelizeOptions ||
+      !options.sequelizeOptions.database ||
+      !options.sequelizeOptions.username ||
+      !options.sequelizeOptions.password
+    )
+      throw new Error(
+        `Cannot use ${dialect} with socketPath without database, username, or password`
+      );
```
patrickhulce, line 104:
```suggestion
    ) {
      throw new Error(
        `Cannot use ${dialect} with socketPath without database, username, or password`
      );
    }
```

MaximeHeckel:
@patrickhulce thank you for the suggestions! Will address the comments and add tests

packages/server/src/api/storage/sql/sql.js
MaximeHeckel: What is the suggestion here? (I can't see the difference between what I added and what you suggested @patrickhulce )

patrickhulce: it's just adding braces around the throw :)

MaximeHeckel:
Running into a hick up for the testing locally (maybe CI will pass?)
Installed and configured MySQL as stated in the docs but getting

```
 FAIL  packages/server/test/mysql-socket-path-server.test.js
  ● mysql server (socket path) › /version › should return the version

    SequelizeHostNotFoundError: getaddrinfo ENOTFOUND /tmp/mysql.sock /tmp/mysql.sock:3306
```

despite the socket being in the `/tmp` folder not sure if that rings a bell to you @patrickhulce I might have messed up my local config

MaximeHeckel:
@patrickhulce maybe you have a bit more context regarding how to use the `mysql` service with Github Actions but I currently don't seem to have permissions to connect through socket path in CI / Linux:
`SequelizeConnectionError: connect EACCES /var/lib/mysql/mysql.sock`

patrickhulce:
> maybe you have a bit more context regarding how to use the mysql service with Github Actions but I currently don't seem to have permissions to connect through socket path in CI / Linux

Well we'll need to start another copy of mysql that exposes the socket path, right now it's just running in a docker container on port 33306, so that socket file won't exist.

Looks like it's not too bad though https://superuser.com/questions/1411402/how-to-expose-linux-socket-file-from-docker-container-mysql-mariadb-etc-to

So steps... 
1. We create a copy of the `mysql` entry currently under services but have it use a socket instead. 
2. Expose the socket folder to the host image like the answer linked above
3. Do the same steps for creating our test database like we do for the port-based on but using the host socket file.
4. Set the environment variable like you do now to get the tests to use this socket file.

.github/workflows/ci.yml
```diff
@@ -74,12 +89,17 @@ jobs:
           echo "::set-env name=CHROME_PATH::$(which google-chrome-stable)"
           echo "::set-env name=POSTGRES_DB_URL::postgres://postgres:postgres@localhost/lighthouse_ci_test"
           echo "::set-env name=MYSQL_DB_URL::mysql://root:mysql@127.0.0.1:33306/lighthouse_ci_test"
+          echo "::set-env name=MYSQL_SOCKET_PATH::/mnt/mysqld.sock"
```
patrickhulce, line 92: fyi @MaximeHeckel you might want to create subdirs so the docker mount doesn't try to clobber an existing directory with the mount I don't think setting `/mnt` for both postgres and mysql will work

patrickhulce: sorry this is such a pain to debug 😅

MaximeHeckel: > you might want to create subdirs so the docker mount doesn't try to clobber an existing directory 

Yeah ran into that issue originally with `/tmp`

MaximeHeckel: Is there a way wit gh actions to create the directories before the service starts?

patrickhulce: Hm I'm not sure to be honest. I would assume there should be but I haven't tried before

MaximeHeckel: Did a bit of investigation yesterday night, it doesn't look like it's possible sadly. (Couldn't find a lot of documentations around volumes, tried to work around it but didn't find any satisfying solution to this)
What we can do maybe is use another path than `/mnt`? Do you have any suggestions? (I tried `/tmp` as well but `yarn` wouldn't be able to successfully run after putting the sockets there)

patrickhulce: have you tried using the workspace dir yet? something like `${{ github.workspace }}/mysql:/var/run/mysqld`? we need it to be predictable and accessible

MaximeHeckel: Just gave it a try, but like in the previous attempt the `mysqlsocket` service doesn't seem to be able to start: https://github.com/GoogleChrome/lighthouse-ci/pull/389/checks?check_run_id=919336503

patrickhulce:
Oof, yeah this is really rough @MaximeHeckel I'm about to give up as well ☹️  https://github.com/GoogleChrome/lighthouse-ci/actions

I'm inclined to just skip the test here at this point since there's not explicit configuration involved if things are working on your end. sockets through docker just seem rough :/

leave the test file in play at least and maybe I'll play around with github actions some more on the subject later. just remove the environment variable for now in the github actions config I guess

patrickhulce:
@googlebot I consent.

MaximeHeckel:
> I'm inclined to just skip the test here at this point since there's not explicit configuration involved if things are working on your end. sockets through docker just seem rough :/

yeah not an easy one (funny thing is that I used to work for Docker, and I remember struggling as much back then, seems that things haven't improved)

>  just remove the environment variable for now in the github actions config I guess

Will do that right now

MaximeHeckel:
> Oh wait you got it working!? That's amazing! Ignore my removal comment if you fixed it I thought we were stuck in failing limbo :)

Oh sorry, I might have caused some confusion here: 
- I got it working locally
- I got it working if mounting a volume in `/mnt` for example (that's where the logs you see in my previous come from, and that's where the run I linked in that same comment points to)

But like we investigated yesterday, we couldn't get it to work by mounting a volume in `/mnt/mysql` or `/mnt/postgres` because we can't create the folder before these services start.

As it stands now this PR has the tests for both mysql and postgres socket path but they are just not running because like you suggested I removed the environment variable for their respective socket paths and thus the tests are skipped.

patrickhulce:
Ah ok gotcha 👍

PR #394, opened 2020-07-31
Add config for Next.js

samrobbins85:
By default, Next.js builds to the folder `out` when the command `next export` is used. I can't think of an alternate use for the out directory beyond builds, so I think it would be useful if this is one of the defaults.


samrobbins85:
@googlebot I signed it!

Issue #250, opened 2020-03-20
Upload URLs map does not work (--uploadUrlMap)


alekseykulikov:
While working on [LHCI Action](https://github.com/treosh/lighthouse-ci-action), I tried to add a support for results comparison for reports stored in Temporary Public Storage. But it seems, that `getUrlMap` always returns `{"success":false}`.

```bash
LHCI_BUILD_CONTEXT__GITHUB_REPO_SLUG=treosh/lighthouse-ci-action lhci upload \
  --target=temporary-public-storage --uploadUrlMap=true
```

Also, [CI Viewer](https://googlechrome.github.io/lighthouse-ci/viewer/) does not support html as a source of LH reports, and I am not sure how this code works in the case of Temporary Storage:
https://github.com/GoogleChrome/lighthouse-ci/blob/master/packages/cli/src/upload/upload.js#L321

Another improvement would be to add branches support beside `master`: https://github.com/GoogleChrome/lighthouse-ci/blob/master/packages/cli/src/upload/upload.js#L303

**Ideal scenario**: in addition to Lighthouse reports in `links.json`, add support for compare URLs with the latest build to a current branch, and `master`.

patrickhulce:
Thanks @alekseykulikov! Tried to respond to each of your points

> But it seems, that getUrlMap always returns {"success":false}.

Were you doing this recently? I recently received lots of suspension emails from GCP that have deleted lots of our storage objects and I'm working through with their support, so I suspect this might be the issue here.

> Also, CI Viewer does not support html as a source of LH reports

What led you to believe this? Yes it does.

> I am not sure how this code works in the case of Temporary Storage:

CI viewer requests those files on load.

https://github.com/GoogleChrome/lighthouse-ci/blob/5e8ef6ad14b4387d75bb9001ddb4f9f2bce0f147/packages/viewer/src/ui/app.jsx#L47-L71

> Another improvement would be to add branches support beside master:

Yep, see #82 .

>  in addition to Lighthouse reports in links.json, add support for compare URLs with the latest build to a current branch, and master.

I'm not sure what you mean by this. Could you elaborate?

alekseykulikov:
Hey @patrickhulce, thank you for the reply.
Yes, I tried it a week ago. I retried it now, and it worked, which is quite awesome! 

> What led you to believe this?

[...]

PR #402, opened 2020-08-05
feat(cli): log attempts to save url map

chosak:
Related to #250.

### DIFF
--- a/packages/cli/src/upload/upload.js
+++ b/packages/cli/src/upload/upload.js
@@ -343,19 +343,31 @@ async function getPreviousUrlMap(options) {
  */
 async function writeUrlMapToApi(urlMap) {
   const slug = getGitHubRepoSlug();
-  if (!slug) return;
+
+  if (!slug) {
+    print(`No GitHub repository slug found, skipping URL map upload.\n`);
+    return;
+  }
+  print(`Saving URL map for GitHub repository ${slug}...`);
 
   try {
     /** @type {Record<string, string>} */
     const payload = {slug};
     Array.from(urlMap.entries()).forEach(([k, v]) => (payload[k] = v));
-    await fetch(SAVE_URL_MAP_URL, {
+
+    const response = await fetch(SAVE_URL_MAP_URL, {
       method: 'POST',
       body: JSON.stringify(payload),
       headers: {'content-type': 'application/json'},
     });
+
+    if (response.ok) {
+      print(`success!\n`);
+    } else {
+      print(`failed!\n`);
+    }
   } catch (err) {
-    print(`Failed to save urlMap: ${err.message}`);
+    print(`Failed to save URL map: ${err.message}\n`);
   }
 }
 
--- a/packages/cli/test/autorun-static-dir.test.js
+++ b/packages/cli/test/autorun-static-dir.test.js
@@ -16,12 +16,21 @@ describe('Lighthouse CI autorun CLI', () => {
   const autorunDir = path.join(__dirname, 'fixtures/autorun-static-dir');
 
   it('should run all three steps', async () => {
-    const {stdout, stderr, status} = await runCLI(['autorun', '--collect.numberOfRuns=1'], {
-      cwd: autorunDir,
-      env: {LHCI_NO_LIGHTHOUSERC: undefined},
-    });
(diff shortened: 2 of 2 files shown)


packages/cli/src/upload/upload.js
```diff
@@ -342,20 +342,33 @@ async function getPreviousUrlMap(options) {
  * @return {Promise<void>}
  */
 async function writeUrlMapToApi(urlMap) {
+  print(`Saving URL map to temporary public storage...\n`);
```
patrickhulce, line 345: this feels a bit excessive considering we don't even print that much for the github intergration which is much more important. one-line about its status feels enough :)
```suggestion
```

packages/cli/src/upload/upload.js
```diff
@@ -342,20 +342,33 @@ async function getPreviousUrlMap(options) {
...
   const slug = getGitHubRepoSlug();
-  if (!slug) return;
+
+  if (!slug) {
+    print(`No GitHub repository slug found, skipping.\n`);
```
patrickhulce, line 349:
```suggestion
    print(`No GitHub repository slug found, skipping URL map upload.\n`);
```

packages/cli/src/upload/upload.js
```diff
@@ -342,20 +342,33 @@ async function getPreviousUrlMap(options) {
...
+    return;
+  }
+  print(`Saving URL map under GitHub repository slug ${slug}...`);
```
patrickhulce, line 352:
```suggestion
  print(`Saving URL map for GitHub repository ${slug}...`);
```

WDYT? :)

packages/cli/src/upload/upload.js
```diff
@@ -342,20 +342,33 @@ async function getPreviousUrlMap(options) {
...
     });
+
+    if (response.ok) {
+      print(`success!\n`);
+    } else {
+      print(`failed!\n`);
+    }
   } catch (err) {
-    print(`Failed to save urlMap: ${err.message}`);
+    print(`Failed to save URL map: ${err.message}`);
```
patrickhulce, line 371:
```suggestion
    print(`Failed to save URL map: ${err.message}\n`);
```

missed this last time thanks for the fix!

chosak:
I see why the tests are failing; because I've configured them to check for output from an upload, but that doesn't happen on when GHA runs on non-`master` checks. @patrickhulce what's optimal behavior here -- to explicitly disable the upload (and roll back test changes), or maybe hardcode an explicit slug to use in the environment [here](https://github.com/GoogleChrome/lighthouse-ci/blob/1bf2e963766214111787d616ef45f198912cfb79/packages/cli/test/autorun-static-dir.test.js#L21)?

patrickhulce:
> maybe hardcode an explicit slug to use in the environment here?

That seems like the best path forward, hardcode the use of `GoogleChrome/lighthouse-ci-unit-tests` as the slug and force the `--uploadUrlMap` flag?

chosak:
Thanks for the text suggestions, I've fixed and rebased. I'll address your slug suggestion.

Aside: While doing that I noticed that some uses of "GitHub" (for example [`getGitHubRepoSlug`](https://github.com/GoogleChrome/lighthouse-ci/blob/1bf2e963766214111787d616ef45f198912cfb79/packages/utils/src/build-context.js#L315), or the logging statements I've added)  are perhaps overly specific, as one might potentially be running Lighthouse against repositories hosted on other source code providers. But maybe that's outside the scope of this PR.

patrickhulce:
Those names are explicit because they're used for things that are github support only. We don't support any other providers and I consider it a bug that they can pickup anything other than github slugs :)

chosak:
Thanks @patrickhulce!

Interesting side note I discovered while working on the tests -- some of [these](https://github.com/GoogleChrome/lighthouse-ci/blob/1bf2e963766214111787d616ef45f198912cfb79/packages/utils/test/build-context.test.js#L174) tests won't pass locally if your Git `origin` remote doesn't point to this repository (for example, I often use `upstream` for a project and then `origin` for my fork).

patrickhulce:
> some of these tests won't pass locally if your Git origin remote doesn't point to this repository (for example, I often use upstream for a project and then origin for my fork).

Ah yeah thanks for the note I'll loosen those! :)

PR #432, opened 2020-08-27
Update "target" configuration documentation

amishshah:
The current documentation lists only `"lhci", "temporary-public-storage"``as targets, however it seems that "filesystem" is also an option. This pull request adds it to the list.


amishshah:
@googlebot I signed it!

patrickhulce:
thanks @amishshah!

Issue #430, opened 2020-08-27
Desktop performance score lower in CLI than when testing in the browser


amannn:
I'm using `@lhci/cli` to test the desktop viewport of a web app I'm developing. I'm using the `emulatedFormFactor: 'desktop'` setting.

When running in `lhci`, I get this score:

![Screenshot 2020-08-27 at 11 43 43](https://user-images.githubusercontent.com/4038316/91424890-ad467d00-e85a-11ea-81fc-e6506366ae50.png)

However when I run the test with the same exact URL in Chrome, I get this score:

![Screenshot 2020-08-27 at 11 59 47](https://user-images.githubusercontent.com/4038316/91426445-d49e4980-e85c-11ea-8e21-7b453b913901.png)

I have no idea what is causing this.

Note that the mobile performance score of the same website is 100 in `lhci` – exactly the same as in Google Chrome.

Do you have any pointers? The app is private, therefore I sadly can't share a reproduction at this point. Thank you for your help!

Versions:
- @lhci/cli: 0.4.4 (although I got the same result when trying out 0.5.0)
 - Chrome: Version 84.0.4147.135 (Official Build) (64-bit)
 - OS: macOS 10.15.4 (19E287)
 - Chrome Lighthouse: 6.0.0
 - @lhci/cli Lighthouse: 6.0.0

patrickhulce:
Thanks for filing @amannn! 

When you're running from the CLI (like LHCI does) form factor and throttling are controlled separately unlike Chrome DevTools which handles this behind the scenes for you (see [desktop throttling profile](https://github.com/GoogleChrome/lighthouse/blob/c2b7fcb99cf83fe4088f2ce5d40f6ae5724233d6/lighthouse-core/config/constants.js#L41-L46)).

```js
module.exports = {
  ci: {
    collect: {
      settings: {
        emulatedFormFactor: 'desktop',
        throttling: {rttMs: 40, throughputKbps: 10 * 1024, cpuSlowdownMultiplier: 1},
      }
  }
}
```

We'd welcome adding this example config to the docs if you're up for a PR :)

amannn:
Oh, I see! Thanks for your quick reply – now this works as expected 🙂 .

Would it make sense for `lighthouse-ci` to include an option to set the device like lighthouse does it in the Chrome DevTools?

[...]

PR #433, opened 2020-08-27
docs(configuration): Add desktop docs

amannn:
Resolves https://github.com/GoogleChrome/lighthouse-ci/issues/430

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -374,27 +374,6 @@ The [Lighthouse settings object](https://github.com/GoogleChrome/lighthouse/blob
 }
 ```
 
-You can use the `settings` property to configure Lighthouse for testing your site on an emulated desktop device:
-
-```jsonc
-{
-  "ci": {
-    "collect": {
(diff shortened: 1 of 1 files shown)


amannn:
@googlebot I signed it!

docs/configuration.md
```diff
@@ -374,6 +374,27 @@ The [Lighthouse settings object](https://github.com/GoogleChrome/lighthouse/blob
 }
 ```
 
+You can use the `settings` property to configure Lighthouse for testing your site on an emulated desktop device:
```
patrickhulce, line 377: WDYT of putting this down below in the [Common Examples section](https://github.com/GoogleChrome/lighthouse-ci/blob/master/docs/configuration.md#common-examples) where we have a longer list of examples? 

I worry that including too many inline might detract from the primary goal of explaining what each LHCI property does.

amannn: Oh right, I didn't see that section previously. I've moved it now!

PR #434, opened 2020-08-27
Add flag to set the limit of files to be tested.

### LINKED ISSUE #95: Only files in staticDistDir root are collected
Using `autorun` against a static site, only the files in the root of the directory are tested.

e.g. If I have a static site with the following structure:
```
.
|____news
| |____articles.html
|__index.html
```

then only `index.html` is collected.

chronark:
This would fix #95 

Right now lhci will only test 5 files.
With this PR you can specify the limit to any number or even remove it.

Example:
```
lhci autorun                            -> Will test 5 files
lhci autorun --collect.numberOfFiles=10 -> Will test 10 files
lhci autorun --collect.numberOfFiles=0  -> Will test all files
```

Default is still at 5.
A value less or equal than 0 will disable the limit alltogether.

I didn't write any tests yet, because I never wrote CLI tests before, but I'll get around to it soon.
I just wanted to create a PR first to see if you would even consider adding it.


chronark:
@googlebot I signed it!

packages/cli/src/collect/collect.js
```diff
@@ -80,6 +80,12 @@ function buildCommand(yargs) {
       default: 3,
       type: 'number',
     },
+    numberOfFiles: {
+      alias: 'f',
```
patrickhulce, line 84: `-f` feels a little too prominent for a fallback option we don't expect many folks to use, mind removing the alias?

packages/cli/src/collect/collect.js
```diff
@@ -80,6 +80,12 @@ function buildCommand(yargs) {
...
+      description: 'The number of files to test. Disable limit by setting this to 0',
```
patrickhulce, line 85:
```suggestion
      description: 'The maximum number of pages to collect when using the staticDistDir option with no specified URLs. Disable this limit by setting to 0.',
```

maybe?

packages/cli/src/collect/collect.js
```diff
@@ -80,6 +80,12 @@ function buildCommand(yargs) {
       default: 3,
       type: 'number',
     },
+    numberOfFiles: {
```
patrickhulce, line 83: WDYT about...

`maxAutodiscoverUrls`
`maxStaticDistDirPages`
`maxNumberOfPages`

any of those sound good to you? anything that conveys the narrower scope (just applies to static dist dir automatic detection fallback URLs, not URLs or files in general) sgtm

packages/cli/src/collect/collect.js
```diff
@@ -167,7 +173,11 @@ async function startServerAndDetermineUrls(options) {
 
   const urls = urlsAsArray;
   if (!urls.length) {
-    urls.push(...server.getAvailableUrls().slice(0, 5));
+    if (typeof options.numberOfFiles !== 'undefined' && options.numberOfFiles <= 0) {
+      urls.push(...server.getAvailableUrls());
+    } else {
+      urls.push(...server.getAvailableUrls().slice(0, options.numberOfFiles));
+    }
```
patrickhulce, line 180: yargs should always default this to a number for us, WDYT about 

```suggestion
    const maxNumberOfUrls = options.numberOfFiles || Infinity;
    urls.push(...server.getAvailableUrls().slice(0, maxNumberOfUrls));
```

packages/cli/src/collect/collect.js
chronark: Yeah you are right about this one.

packages/cli/src/collect/collect.js
chronark: Thanks that reads much better, I wasn't sure how long the description may be.

packages/cli/src/collect/collect.js
chronark: I think maxAutodiscoverUrls is best,
the other ones could be confusing because you technically can add more urls from the staticDistDir manually,

packages/cli/src/collect/collect.js
chronark: Oh cool I didn't know you could just pass Infinity to slice and make it work this way.
That's a lot cleaner than my if else block. Thanks.

chronark:
Alright here we go. Thank you for guiding me @patrickhulce and for your suggestions, I learned something new :)

I added some more tests, so we are testing if it correctly 
- limits it to a number
- disables the limit
- defaults to 5

I had to create a new fixture folder but I hope that is fine.

I also added the flag to the configuration.md

Issue #444, opened 2020-09-16
`lhci` incorrectly tries to serve (and audit) folder with `.html` extension


ojizero:
We are experiencing an issue where one of our Gatsby generated output, is a folder ending with `.html` in the name, this is prompting `lhci collect` step to try and serve it as if it is an HTML page even though it's just a normal folder.

The server code for reference where the filtering is only done on the file or folder's name only, specifically line 88. https://github.com/GoogleChrome/lighthouse-ci/blob/c353ee8937bc0ca1474905baeb8d53799449ec56/packages/cli/src/collect/fallback-server.js#L85-L89

This is causing failures as Lighthouse tries to serve it and fails to do so since it's a folder and not a page.

Is this an expected behaviour, and if so is there any workaround for it, or can it be considered a buggy behaviour on Lighthouse CI CLI part?

patrickhulce:
Yeah we can consider this a bug. I would've expected a folder ending in `.html` to end up serving some content via an `index.html` *within* that folder, but we probably shouldn't autoselect it either way. It's a pretty easy fix if anyone is interested :) We can use `fs.readdirSync(directory, {withFileTypes: true})` to filter out directories in that line you flagged.

The workaround is to explicitly specify what URLs you want Lighthouse CI to collect.

`lhci collect --url=http://localhost/what/to/audit --url=http://localhost/another/url`

ojizero:
@patrickhulce thanks for the quick response! just opened PR for it #445

PR #445, opened 2020-09-16
Serve only files when running `lhci collect`

ojizero:
Fixes the behaviour where the FallbackServer attempts to serve folders that
happen to end with `.html` causing errors. Filters scanned directory
for only files in it before executing any of the logic there.

Fixes: #444


patrickhulce (commented):
awesome thank you @ojizero! would you be able to add this to one of our autorun test cases? 

I'm thinking it's fine if you just add a file `folder.html/invisible.html` to https://github.com/GoogleChrome/lighthouse-ci/tree/master/packages/cli/test/fixtures/autorun-static-dir/build  and make sure we don't pick it up in the existing test, no need to write a new one

ojizero:
@patrickhulce hmmmm 🤔 just thinking out loud here but isn't the condition `if (folder.includes('.')) continue;` kind of an overkill? maybe the URL does have `.` in it and has other things living under it? aren't those valid pages to check on?

shouldn't the check be only for hidden files? i.e. `something.other/somepage.html` should be a valid page that gets audited? maybe this can be treated in another PR but I'm just thinking out loud here since I think the suggested `folder.html/invisible.html` should in fact be visible given that the depth being checked is `2` 🤔

patrickhulce:
> isn't the condition if (folder.includes('.')) continue; kind of an overkill? maybe the URL does have . in it and has other things living under it? aren't those valid pages to check on?

Maybe, but I don't think we should change it here for two reasons.

1) This is just for automatic defaults, it's not restrictive of what you can do. A folder with a dot is at the very least unusual and I'm OK with those URLs needing explicit opt-in.
2) Any change that automatically starts including *more* files here would need to be in a major breaking version per [version policy](https://github.com/GoogleChrome/lighthouse-ci/blob/master/docs/version-policy.md).

ojizero:
very true yea changing this would be a breaking change and may introduce unexpected behaviours ! I'll add the new fixture for the test as it is now 👍

packages/cli/test/fixtures/autorun-static-dir/build/folder.html/invesible.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
patrickhulce, line 1: can you rename this file `invisible.html`? :) just spelling typo

patrickhulce (approved):
implementation looks great to me, thanks @ojizero! 🎉

packages/cli/test/fixtures/autorun-static-dir/build/folder.html/invesible.html
ojizero: if I got a dollar for every typo I make 💃

PR #449, opened 2020-09-17
Update Docs on SQL Connection URL Format

tsvanharen:
Under the covers, it seems that the value of `sqlConnectionUrl` is passed through to Sequelize, so we should be more clear about what's allowed in the format of that string.  Relevant Sequelize documentation on the proper format can be seen here:  https://sequelize.org/v4/manual/installation/getting-started.html#setting-up-a-connection.


patrickhulce (approved):
thanks @tsvanharen! 🎉

PR #457, opened 2020-09-23
feat(server): add compare different url

pndllxzzy:


pndllxzzy:
@googlebot I signed it!

patrickhulce:
thanks for the PR @pndllxzzy! It gets a little crowded by sticking it up there though and the click target moves when you hover because of "Open Report" :/
image


Other ideas:

 - hide the base and compare URL selectors behind the menu where you select the build/hash
 - have both dropdowns in the original location of the compare URL dropdown (add more labels to differentiate)

pndllxzzy:
Thanks for your ideas.Do you mean like this?
image

patrickhulce:
Yeah, that seems to get the job done even if it's not the prettiest. WDYT?

pndllxzzy:
@patrickhulce I agree with you. I'm not good at beautiful layout. If you have any ideas, please tell me.🤝🏻

packages/server/src/ui/routes/build-view/lhr-comparison.jsx
```diff
@@ -113,13 +113,14 @@ export function computeAuditGroups(lhr, baseLhr, options) {
   return auditGroups;
 }
 
-/** @param {{hookElements: LHCI.HookElements<'dropdowns'>, selectedAuditId?: string | null, lhr?: LH.Result, baseLhr?: LH.Result, percentAbsoluteDeltaThreshold: number, setPercentAbsoluteDeltaThreshold: (x: number) => void}} props */
```
patrickhulce, line 116: I think we can leave this just as `dropdowns` it was already set to be plural for more than one. 

use a fragment when it's used instead?

packages/server/src/ui/routes/build-view/build-view.jsx
```diff
@@ -126,7 +130,7 @@ const BuildView_ = props => {
             lhr={lhr}
             isDimmed={openBuildHash === 'base'}
             isOpen={openBuildHash === 'compare'}
-            setLhrLinkHover={setLhrLinkHover}
+            setLhrLinkHover={setLhrCompareLinkHover}
```
patrickhulce, line 133: nice split 👍

packages/server/src/ui/routes/build-view/build-view.jsx
```diff
@@ -159,24 +163,42 @@ const BuildView_ = props => {
...
         hookElements={{
           warnings: computeWarnings(warningProps).hasWarning ? (
             <BuildViewWarnings {...warningProps} />
           ) : (
             undefined
           ),
-          dropdowns: (
+          dropdownsBase: (
+            <Dropdown
+              label="BaseUrl"
```
patrickhulce, line 178:
```suggestion
              label="Base URL"
```

packages/server/src/ui/routes/build-view/build-view.jsx
```diff
@@ -159,24 +163,42 @@ const BuildView_ = props => {
...
+              }}
+              options={availableUrlOptions}
+            />
+          ),
+          dropdownsCompare: (
             <Dropdown
-              label="URL"
-              className="dropdown--url"
-              value={selectedUrl}
+              label="CompareUrl"
```
patrickhulce, line 192:
```suggestion
              label="Compare URL"
```

patrickhulce (commented):
thanks @pndllxzzy this is great! I think you might need to update some snapshot tests as well?

packages/server/src/ui/routes/build-view/lhr-comparison.jsx
pndllxzzy: Do you mean like this?
```
<LhrComparison
  ...
  hookElements={{
    ...
    dropdowns: (
      <Fragment>
        <Dropdown
          label="Base Url"
          className="dropdown--url dropdown--base-url"
          value={baseUrl}
          setValue={url => {
            const to = new URL(window.location.href);
            to.searchParams.set('baseUrl', url);
            to.searchParams.set('compareUrl', compareUrl);
            route(`${to.pathname}${to.search}`);
          }}
          options={availableUrlOptions}
        />
        <Dropdown
          label="Compare Url"
          className="dropdown--url dropdown--compare-url"
          value={compareUrl}
          setValue={url => {
            const to = new URL(window.location.href);
            to.searchParams.set('baseUrl', baseUrl);
            to.searchParams.set('compareUrl', url);
            route(`${to.pathname}${to.search}`);
          }}
          options={availableUrlOptions}
        />
      </Fragment>
    ),
  }}
/>
```

pndllxzzy:
@patrickhulce How to update snapshot?
Use this command?
`jest --updateSnapshot`

patrickhulce:
Yep! I tend to use `-u` but that works too :)

patrickhulce (changes requested):
just these two leftover from previous review but other than that looks good with test fixes! :) thanks for doing this @pndllxzzy

pndllxzzy:
@patrickhulce Should I rebase all commits to one?

patrickhulce:
> Should I rebase all commits to one?

Don't worry about it we'll squash when we merge :) I'll take a closer look here tomorrow 👍

patrickhulce (approved):
thanks for sticking with this @pndllxzzy! 🎉

PR #468, opened 2020-10-13
Add flag to filter out URLs you don't want to be tested

JohnEidum:
This PR allows you to filter out certain URLs that you would like to be skipped over during testing. This is used alongside staticDistDir and is not functional with url (Not sure how exactly that would end up working).

### DIFF
--- a/packages/cli/src/collect/collect.js
+++ b/packages/cli/src/collect/collect.js
@@ -191,7 +191,7 @@ async function startServerAndDetermineUrls(options) {
         return url.href;
       }
     );
-    const urlsToUse = availableUrls.filter(url => !normalizedBlocklist.includes(url)).slice(0, maxNumberOfUrls)
+    const urlsToUse = availableUrls.filter(url => !normalizedBlocklist.includes(url)).slice(0, maxNumberOfUrls);
     urls.push(...urlsToUse);
   }
 
--- a/packages/cli/src/collect/collect.js
+++ b/packages/cli/src/collect/collect.js
@@ -184,14 +184,14 @@ async function startServerAndDetermineUrls(options) {
       ? [options.autodiscoverUrlBlocklist]
       : [];
     const availableUrls = server.getAvailableUrls();
-    const normalizedBlocklist = autodiscoverUrlBlocklistAsArray.map(
-      rawUrl => {
-        const url = new URL(rawUrl, 'http://localhost');
-        url.port = server.port.toString();
-        return url.href;
-      }
-    );
-    const urlsToUse = availableUrls.filter(url => !normalizedBlocklist.includes(url)).slice(0, maxNumberOfUrls);
+    const normalizedBlocklist = autodiscoverUrlBlocklistAsArray.map(rawUrl => {
+      const url = new URL(rawUrl, 'http://localhost');
+      url.port = server.port.toString();
+      return url.href;
+    });
+    const urlsToUse = availableUrls
+      .filter(url => !normalizedBlocklist.includes(url))
+      .slice(0, maxNumberOfUrls);
     urls.push(...urlsToUse);
   }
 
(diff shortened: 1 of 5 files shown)


JohnEidum:
@googlebot I signed it!

JohnEidum:
@googlebot I fixed it

JohnEidum:
@googlebot I fixed it.

packages/cli/src/collect/collect.js
```diff
@@ -39,6 +39,10 @@ function buildCommand(yargs) {
       description:
         'A URL to run Lighthouse on. Use this flag multiple times to evaluate multiple URLs.',
     },
+    filterUrl: {
```
patrickhulce, line 42: WDYT about... 

`ignoreAutodiscoverUrl`
`skipAutodiscoverUrl`
`autodiscoverUrlBlocklist`
`autodiscoverSkipUrl`
`autodiscoverIgnoreUrl`

`skipAutodiscoverUrl` and `autodiscoverUrlBlocklist` are probably my personal favorites.

rationale:

- `autodiscover` feels like it needs to be in there somewhere to signify it only works for autodiscovered URLs (comment isn't totally accurate atm, since manually specified URLs can still be used with `staticDistDir`)
- `filterUrl` makes me think it's a regex pattern for *allowing* certain URLs through, `ignore`/`skips`/`block` are more explicit

packages/cli/src/collect/collect.js
```diff
@@ -174,7 +178,13 @@ async function startServerAndDetermineUrls(options) {
...
   if (!urls.length) {
     const maxNumberOfUrls = options.maxAutodiscoverUrls || Infinity;
-    urls.push(...server.getAvailableUrls().slice(0, maxNumberOfUrls));
+    const filterUrlsAsArray = Array.isArray(options.filterUrl)
+      ? options.filterUrl
+      : options.filterUrl
+      ? [options.filterUrl]
+      : [];
+    urls.push(...server.getAvailableUrls().filter(url => 
+      !filterUrlsAsArray.some(pattern => url.endsWith(":" + server.port + pattern))).slice(0, maxNumberOfUrls));
```
patrickhulce, line 187: let's split this out into some components rather than all one big line

```js
const availableUrls = server.getAvailableUrls();
const normalizedBlocklist = filterUrlsAsArray.map(...)
const urlsToUse = availableUrls.filter(url => !normalizedBlocklist.includes(url)).slice(0, maxNumberOfUrls)
```

> url.endsWith(":" + server.port + pattern))

we should reuse our logic below on 195-197 for consistency, not just appending port/pattern :)

patrickhulce (changes requested):
thanks very much @JohnEidum! this functionality sounds good, just some naming and code clarity points :)

JohnEidum:
Thank you! I have made the necessary changes. I chose to go with `autodiscoverUrlBlocklist`.

patrickhulce (approved):
thanks @JohnEidum 🎉

packages/cli/src/collect/collect.js
```diff
@@ -174,7 +178,21 @@ async function startServerAndDetermineUrls(options) {
...
+      : [];
+    const availableUrls = server.getAvailableUrls();
+    const normalizedBlocklist = autodiscoverUrlBlocklistAsArray.map(
+      rawUrl => {
+        const url = new URL(rawUrl, 'http://localhost');
+        url.port = server.port.toString();
+        return url.href;
+      }
+    );
+    const urlsToUse = availableUrls.filter(url => !normalizedBlocklist.includes(url)).slice(0, maxNumberOfUrls)
```
patrickhulce, line 194:
```suggestion
    const urlsToUse = availableUrls.filter(url => !normalizedBlocklist.includes(url)).slice(0, maxNumberOfUrls);
```

I suspect this line might fail lint FYI

patrickhulce:
sorry @JohnEidum you have one more annoying snapshot to update with help text whitespace padding

Issue #172, opened 2020-01-07
Not seeing status checks despite passing LHCI_GITHUB_APP_TOKEN env


klapec:
Hi,

I'm trying to run lighthouse-ci as part of my Github Actions workflow. I've got a following step defined:
```yaml
- name: Run Lighthouse CI
  run: |
    npm install -g @lhci/cli@0.3.x
    sh ./scripts/lighthouse-ci.sh
  env:
    LHCI_GITHUB_APP_TOKEN: ${{ secrets.LHCI_GITHUB_APP_TOKEN }}
```
and here's the lighthouse-ci script:
```bash
#!/bin/bash

yarn serve-ssr & npx wait-on http://localhost:3000/ping && lhci autorun \
  --upload.target=temporary-public-storage \
  --collect.url="http://localhost:3000/" \
  --collect.numberOfRuns=3

EXIT_CODE=$?
exit $EXIT_CODE
```
The job finishes successfully, and yet there are no status checks from lhci app on the pull request pages.
The Lighthouse CI GitHub app is installed.
Here's the lhci output:
```
2020-01-07T11:08:08.1615866Z ##[group]Run npm install -g @lhci/cli@0.3.x
2020-01-07T11:08:08.1616571Z [36;1mnpm install -g @lhci/cli@0.3.x[0m
2020-01-07T11:08:08.1616905Z [36;1msh ./scripts/lighthouse-ci.sh[0m
2020-01-07T11:08:08.1650457Z shell: /bin/bash -e {0}
2020-01-07T11:08:08.1651111Z env:
2020-01-07T11:08:08.1651413Z   SSH_AUTH_SOCK: /tmp/ssh-auth.sock
2020-01-07T11:08:08.1654217Z   LHCI_GITHUB_APP_TOKEN: ***
2020-01-07T11:08:08.1654571Z ##[endgroup]
2020-01-07T11:08:20.6725461Z /opt/hostedtoolcache/node/12.13.0/x64/bin/lhci -> /opt/hostedtoolcache/node/12.13.0/x64/lib/node_modules/@lhci/cli/src/cli.js
2020-01-07T11:08:20.7986075Z + @lhci/cli@0.3.7
2020-01-07T11:08:20.7987248Z added 372 packages from 265 contributors in 12.174s
2020-01-07T11:08:21.2479204Z yarn run v1.19.2
2020-01-07T11:08:21.2922368Z $ node ./server/bootstrap.js
[...]

PR #484, opened 2020-11-04
docs: Improve GitHub Actions snippets

carloscuesta:
Hello! 👋🏼 

Yesterday I was migrating from Travis-CI to GitHub Actions and I found a couple of things that are not mentioned on the docs that could be helpful to include

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,14 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-        with:
-          ref: ${{ github.event.pull_request.head.sha }}
       - uses: actions/setup-node@v1
       - run: npm install && npm install -g @lhci/cli@0.6.x
       - run: npm run build
       - run: lhci autorun
-        env:
-          LHCI_GITHUB_APP_TOKEN: ${{ secrets.LHCI_GITHUB_APP_TOKEN }}
 ```
 
 ### Features
--- a/README.md
+++ b/README.md
@@ -18,10 +18,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,10 +18,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
carloscuesta, line 22: This is related with the following issue: https://github.com/GoogleChrome/lighthouse-ci/issues/172

README.md
```diff
@@ -18,10 +18,14 @@ jobs:
...
       - uses: actions/setup-node@v1
       - run: npm install && npm install -g @lhci/cli@0.6.x
       - run: npm run build
       - run: lhci autorun
+        env:
+          LHCI_GITHUB_APP_TOKEN: ${{ secrets.LHCI_GITHUB_APP_TOKEN }}
```
carloscuesta, line 28: This is mentioned on the docs, that you should add the LHCI_GITHUB_APP_TOKEN as an environment variable on the repository but you should include as env inside of the action

patrickhulce (changes requested):
Thanks for the feedback and improvements @carloscuesta! 🎉 

These two changes are both specific to GitHub status checks and we've tried quite intentionally to avoid including too many future steps in the early parts of the documentation (like the quick start and step 1 of getting started that are changed here.

What do you think of adding this instead under "GitHub Status Checks" section with a special note about ensuring git history is available in *all* CI providers and providing the example for GitHub actions like you've done?

carloscuesta:
@patrickhulce I think it makes sense! I'm making the updates you suggested, thanks for taking the time to review the PR 😊

patrickhulce (approved):
thanks @carloscuesta!

carloscuesta:
You're welcome! Happy to contribute 😊

PR #501, opened 2020-11-29
Adding in Cloudbuild documentation

annez:
Thought it'd be good to add in Google Cloudbuild. I've also upgraded the referred image on GitLab as that image doesn't seem to work with Lighthouse 6.

We're running on HIGHCPU_32 as it delivers the best and most consistent results (and HIGHCPU_8) is similar to Github Actions which doesn't return consistent results.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -168,7 +168,7 @@ module.exports = {
 **.gitlab-ci.yml**
 
 ```yaml
-image: cypress/browsers:node10.16.0-chrome77
+image: cypress/browsers:node14.15.0-chrome86-ff82
 lhci:
   script:
     - npm install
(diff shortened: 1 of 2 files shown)


annez:
@googlebot I signed it!

docs/getting-started.md
```diff
@@ -221,6 +221,58 @@ lhci autorun
...
+    upload: {
+      target: 'temporary-public-storage',
+    },
+  },
+};
+```
+
+**Notes**
+* `LHCI_BUILD_CONTEXT__CURRENT_BRANCH` doesn't pick up the right variables in cloudbuild so passing through `$BRANCH_NAME` will fix this.
+* `machineType` defines the machine you can pick. The bigger the better (see: https://github.com/GoogleChrome/lighthouse/blob/master/docs/variability.md). Look through the [Cloudbuild machine pricing](https://cloud.google.com/cloud-build/pricing)
```
patrickhulce, line 247:
```suggestion
* `machineType` defines the machine you can pick. The bigger the machine the more stable the performance results will be (see [Lighthouse variability documentation](https://github.com/GoogleChrome/lighthouse/blob/master/docs/variability.md)). WARNING: Look through the [Cloudbuild machine pricing](https://cloud.google.com/cloud-build/pricing) before deciding on a machine type.
```

docs/getting-started.md
```diff
@@ -221,6 +221,58 @@ lhci autorun
...
+  - id: 'lighthouse'
+    waitFor: ['build']
+    name: cypress/browsers:node14.15.0-chrome86-ff82
+    entrypoint: '/bin/sh'
+    args: ['-c', 'npm install -g @lhci/cli@0.6.x && lhci autorun --failOnUploadFailure']
+    env:
+      - 'LHCI_BUILD_CONTEXT__CURRENT_BRANCH=$BRANCH_NAME'
+
+options:
+  machineType: 'N1_HIGHCPU_32'
```
patrickhulce, line 271:
```suggestion
  machineType: 'N1_HIGHCPU_8'
```

WDYT about starting with 8 by default? the getting started guide isn't meant for folks to be relying on the stability of metric results from the get-go and maybe we add a note to the later stages about increasing the machine type (for whatever your CI is)? 

that way all the examples are fairly consistent too since I know the other machine types in other CI examples won't have 32 cores :)

patrickhulce (changes requested):
thanks very much for the addition and the missed update to the docker image! :)

docs/getting-started.md
annez: Yeah agreed. We didn't test at that due to our build having many other bits to it. I think good to have it at 8 and mention the ability to improve. 

Happy to take a look at the other CIs and try and document something here

patrickhulce (approved):
thanks for helping out @annez! 🎉

PR #502, opened 2020-11-30
Fix #500 by using uuid to generate the random filename

### LINKED ISSUE #500: Flags file cleanup failure
We have integrated lighthouse in our CI pipeline and run them on a number of URLs. In order to reduce the time for the workflow, we decided to do concurrent runs. We're aware that it might introduce variability and we're fine with that (the instances we run lighthouse on are good enough that we don't observe a lot of variability). The concurrent runs reduced the time by half. However, we're currently encountering the error below:
```
internal/fs/utils.js:454
    throw err;
    ^

Error: ENOENT: no such file or directory, unlink '/root/gaf/webapp/.lighthouseci/flags-2516.json'
    at Object.unlinkSync (fs.js:991:3)
    at cleanupFn (/root/gaf/webapp/node_modules/@lhci/cli/src/collect/node-runner.js:55:28)
    at ChildProcess.<anonymous> (/root/gaf/webapp/node_modules/@lhci/cli/src/collect/node-runner.js:87:7)
    at ChildProcess.emit (events.js:196:13)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:257:12)
```

I am guessing that it might be because of how the filenames are generated. Since the random string generation is simple, it is possible that two processes could watch the same file. When the file gets cleaned up by the first process, the second process will fail. I have not confirmed the issue and this is just a guess so feel free to correct me if I'm wrong. This doesn't happen often but any help is appreciated. Thanks!

https://github.com/GoogleChrome/lighthouse-ci/blob/v0.6.1/packages/cli/src/collect/node-runner.js#L64 
```
      const flagsFilename = `flags-${Math.round(Math.random() * 10000)}.json`;
      const flagsFilePath = path.join(getSavedReportsDirectory(), flagsFilename);
      lhArgs.push('--cli-flags-path', flagsFilePath);

      fs.writeFileSync(flagsFilePath, JSON.stringify(settings));
      cleanupFn = () => fs.unlinkSync(flagsFilePath);
```

PJHRobles:
Fix #500 by using a more robust random filename generation

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -24,6 +24,7 @@
     "open": "^7.1.0",
     "tmp": "^0.1.0",
     "update-notifier": "^3.0.1",
+    "uuid": "^8.3.1",
     "yargs": "^12.0.5",
     "yargs-parser": "^11.1.1"
   }
--- a/packages/cli/src/collect/node-runner.js
+++ b/packages/cli/src/collect/node-runner.js
@@ -8,6 +8,7 @@
 const os = require('os');
 const fs = require('fs');
 const path = require('path');
+const uuid = require('uuid');
 const childProcess = require('child_process');
 const {getSavedReportsDirectory} = require('@lhci/utils/src/saved-reports.js');
 
@@ -61,7 +62,7 @@ class LighthouseRunner {
     const lhArgs = [url, '--output', 'json', '--output-path', 'stdout'];
 
     if (Object.keys(settings).length) {
-      const flagsFilename = `flags-${Math.round(Math.random() * 10000)}.json`;
+      const flagsFilename = `flags-${uuid.v4()}.json`;
       const flagsFilePath = path.join(getSavedReportsDirectory(), flagsFilename);
       lhArgs.push('--cli-flags-path', flagsFilePath);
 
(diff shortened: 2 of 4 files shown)


patrickhulce (changes requested):
thanks very much @PJHRobles this is great! elegant, simple solution 👍

would you mind updating the other dependency in LHCI to match this version requirement? I believe its in the server package.json

rerunning yarn to consolidate would be much appreciated too.

PJHRobles:
Not sure what happened on the mac tests, `STDOUT` and `STDERR` doesn't show anything.
```
FAIL packages/cli/test/upload.test.js (9.152s)
  ● Lighthouse CI upload CLI › should upload for a build

    Server failed to start
    STDOUT: 
    STDERR:

      107 |     await waitForCondition(() => stdout.includes('listening'), 'Server failed to start');
      108 |   } catch (err) {
    > 109 |     throw new Error(`${err.message}\nSTDOUT: ${stdout}\nSTDERR: ${stderr}`);
          |           ^
      110 |   }
      111 | 
      112 |   const port = stdout.match(/port (\d+)/)[1];
```
Are these perhaps flaky tests?

patrickhulce (commented):
thanks @PJHRobles it's possible it's just a flake, I kicked off a restart. 

Looks like the `@types/uuid` package version and the yarn.lock didn't get updated with the server change, but LGTM other than that! :)

patrickhulce (approved):
thanks @PJHRobles!

PR #505, opened 2020-12-09
Added DRONE_BRANCH to env

mlnchk:


mlnchk:
@googlebot I signed it!

patrickhulce (commented):
thanks very much @mlnchk! shall we add the other `DRONE_*` environment variables?

`DRONE_COMMIT_SHA`
`DRONE_REPO`

in particular, bummer that there's no build URL variable

mlnchk:
@patrickhulce commit SHA already works (don't know why),  and where to use REPO I have not found. But just in case, I'll add them, thanks :)

mlnchk:
build url does not exist because Drone is a self-hosted solution I think :(

patrickhulce (approved):
Thanks @mlnchk !

PR #470, opened 2020-10-14
feat(cli): upgrade to Lighthouse 6.4.1

csabapalfi:
BREAKING CHANGE: new audits

fixes #443

PR #513, opened 2020-12-16
Upgrade upstream lighthouse dependency to version 6.5.0

LukeLeber:
I had noticed a warning while viewing Lighthouse reports lately.

> Warning: Results may not display properly. Report was created with an earlier version of Lighthouse (6.4.1). The latest version is 6.5.0.

It seems that there was an upstream release to a dependency.  Opening a PR to pull in the latest upstream version.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -19,7 +19,7 @@
     "express": "^4.17.1",
     "inquirer": "^6.3.1",
     "isomorphic-fetch": "^2.2.1",
-    "lighthouse": "6.4.1",
+    "lighthouse": "6.5.0",
     "lighthouse-logger": "1.2.0",
     "open": "^7.1.0",
     "tmp": "^0.1.0",


LukeLeber:
@googlebot I signed it!

patrickhulce:
Thanks very much for the contribution @LukeLeber!

Unfortunately for us, upgrading the Lighthouse version is a significantly more involved process (see https://github.com/GoogleChrome/lighthouse-ci/pull/470 and https://github.com/GoogleChrome/lighthouse-ci/pull/416 as examples) so we don't track every single update. With v7 of Lighthouse on the horizon any day now, I'm inclined to wait a few days for that to release instead.

Was there something urgent in 6.5.0 that piqued your interest?

LukeLeber:
Hi @patrickhulce  - nothing in particular.  We're getting ready to roll lighthouseci out as part of our pipeline and it seemed like an easy contribution on the surface.  I wasn't aware that there were additional headaches outside of normal SEMVER ones. :)

PR #520, opened 2020-12-31
readme quick start use the new actions/setup-node@v2

janosh:
The stable release of `actions/setup-node@v2` landed two weeks ago:

https://github.com/actions/setup-node/releases/tag/v2.1.4

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v2
       - run: npm install && npm install -g @lhci/cli@0.6.x
       - run: npm run build
       - run: lhci autorun


janosh:
Actually I just noticed that CI with `actions/setup-node@v2` fails due to `npm install -g` not having permission to install to `/usr/local/lib/node_modules/`. Probably a thing to look into.

patrickhulce:
> Actually I just noticed that CI with actions/setup-node@v2 fails due to npm install -g not having permission to install to /usr/local/lib/node_modules/. Probably a thing to look into.

Ah nice catch. Yeah if that's the case, we'll either stick to v1 or hold off until we see a working example in the wild of this with the correct `sudo` sprinkling :)

patrickhulce:
We're going to go ahead and close this. v1 is serving us well and until we have a compelling reason to move to v2 with validation of it working in the wild we'll continue to hold off.

PR #524, opened 2021-01-06
Update docker client version

ddunkin:


docs/recipes/docker-client/Dockerfile
```diff
@@ -11,8 +11,7 @@ RUN apt-get update \
     && apt-get clean
 
 # Install Lighthouse CI
-RUN npm install -g @lhci/cli@0.6.0
-RUN npm install -g lighthouse
```
patrickhulce, line 15: Why is this one being removed? This client image is frequently used as a regular lighthouse docker image as well, without it I'm not sure `lighthouse` is available in global scope.

patrickhulce (commented):
thanks @ddunkin!

docs/recipes/docker-client/Dockerfile
ddunkin: @lhci/cli will use its own version of lighthouse from its dependencies. The version installed separately may be another version. If you build the image from master now, you'll get lighthouse 7.0.0.

> This client image is frequently used as a regular lighthouse docker image as well

Oh, I didn't know. The documentation only talks about the image used for lhci. Should using the image to run lighthouse directly be documented?

patrickhulce: > @lhci/cli will use its own version of lighthouse from its dependencies. The version installed separately may be another version. If you build the image from master now, you'll get lighthouse 7.0.0.

Right that's the feature part of this I didn't want to lose :) haha

> Oh, I didn't know. The documentation only talks about the image used for lhci. Should using the image to run lighthouse directly be documented?

Yeah a comment describing its purpose sounds good 👍 It wasn't an intentional use case, but it kinda turned into a useful setup people started using and seems harmless to keep it working.

ddunkin: > > @lhci/cli will use its own version of lighthouse from its dependencies. The version installed separately may be another version. If you build the image from master now, you'll get lighthouse 7.0.0.
> 
> Right that's the feature part of this I didn't want to lose :) haha

You want the version of lighthouse installed globally to be different from the one lhci uses?

patrickhulce: If you're asking if it would be preferable that they match, sure! And while we're at it it'd be even better to have a separate `lighthouse` only docker image whose version tags match the actual lighthouse version instead of whatever happened to be latest when this was published. 

Because that pipeline is more work I don't think either of us are jumping up and down to setup at this exact moment ;), the important feature I don't want to lose here that some have come to rely on is that *a* recent version of Lighthouse is available globally, which as you noted isn't exposed by the inner lighthouse dependency that `@lhci/cli` creates.

ddunkin: Alright, I just backed out that change.

patrickhulce (approved):

patrickhulce:
thanks very much @ddunkin! :)

PR #526, opened 2021-01-08
docs(recipes): Update Kubernetes deployment instructions

ddunkin:


patrickhulce (changes requested):
thanks very much @ddunkin! I'll be honest I don't know much about k8s, so your help is much appreciated but my review is not all that valuable :) seems good!

just the one script change

patrickhulce (approved):
thanks!

PR #531, opened 2021-01-12
feat(server): add a health check route before auth

ddunkin:
Google Kubernetes Engine default ingress requires a health check route that returns a 200 response. If basic authentication is enabled, there is no route that can be used for a health check. This PR adds a `/healthz` route before the auth middleware to be used for health checks.

### DIFF
--- a/docs/recipes/docker-server/kubernetes/lhci-deployment.yml
+++ b/docs/recipes/docker-server/kubernetes/lhci-deployment.yml
@@ -26,7 +26,7 @@ spec:
             - containerPort: 9001
           readinessProbe:
             httpGet:
-              path: /app/
+              path: /healthz
               port: 9001
       volumes:
         - name: lhci-data-volume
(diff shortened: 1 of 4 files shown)


patrickhulce:
thanks @ddunkin! where does `/healthz` come from? :) 

- `/ping`
- `/health`
- `/status`

All come to mind as options, but if the `z` has significance I don't know of, happy to leave as-is :)

ddunkin:
`/healthz` is a common convention because it's used in a lot of examples, at least in Kubernetes and Google Cloud. I don't think there's a significance to the `z`; the story I've heard is that long ago some Google internal service needed a health check and `/health` conflicted with an existing endpoint, so they tacked a `z` on the end and it stuck.

patrickhulce (approved):

patrickhulce:
cool, works for me then! a test that this new endpoint works would be much appreciated and we're good to merge :)

patrickhulce (approved):
thanks!

PR #532, opened 2021-01-14
build(deps-cli): fix npm vulnerabilities

Divlo:
When installing `@lhci/cli`, npm show us "4 low severity vulnerabilities" : 
Capture1_43
Capture1_44

We should not use dependencies with vulnerabilities.

In the future to avoid these sort of things, we could use [dependabot](https://dependabot.com/), it is a builtin solution in GitHub to automatically update dependencies.

**Link about the vulnerabilities :**
- https://www.npmjs.com/advisories/1500/versions
- https://www.npmjs.com/advisories/1556

**What changes this PR introduce ?**

- Bump `isomorphic-fetch` from 2.2.1 to 3.0.0
  See: https://github.com/matthew-andrews/isomorphic-fetch/issues/192
  The only things to be aware by doing this upgrade : "Node.js 0.10.x and 0.12.x support is dropped"
- Bump `yargs` from 12.0.5 to 15.4.1
  Didn't check what need to change in the codebase to support the latest version of `yargs` but hopefully the `ci` pass and 
  everything is fine, if not we would need to see what's the breaking changes in the few releases of `yargs` and update the code 
  base accordingly.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -18,14 +18,14 @@
     "compression": "^1.7.4",
     "express": "^4.17.1",
     "inquirer": "^6.3.1",
-    "isomorphic-fetch": "^2.2.1",
+    "isomorphic-fetch": "^3.0.0",
     "lighthouse": "6.4.1",
     "lighthouse-logger": "1.2.0",
     "open": "^7.1.0",
     "tmp": "^0.1.0",
     "update-notifier": "^3.0.1",
     "uuid": "^8.3.1",
-    "yargs": "^12.0.5",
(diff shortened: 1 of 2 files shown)


patrickhulce (approved):
thanks very much @Divlo!

Divlo:
With pleasure, as a user, I find out annonying to have vulnerabilities. :smile: 
Excuse me for running your `CI` that much, I forget to also update `isomorphic-fetch` in `@lhci/utils`, we should release a new version of this package, and update the version where it is used (I can't do that in this PR, since we first need to release a new version). @patrickhulce

PR #538, opened 2021-01-27
feat(cli): enable CLI args for lhci wizard new-project

revolunet:
Hi,

This makes it possible to run the `new-project` command without user interaction.

One could create a project like this :

```sh
lhci wizard --wizard=new-project \
  --basicAuth.username=admin \
  --basicAuth.password=io \
  --serverBaseUrl="https://lighthouse.hello.world" \
  --projectName="some-new-project" \
  --projectExternalUrl="https://somewhere.else" \
  --projectBaseBranch="master"

```


revolunet:
@googlebot I fixed it.

packages/cli/src/wizard/wizard.js
```diff
@@ -29,33 +29,46 @@ async function runNewProjectWizard(options) {
   const fallbackProjectName = path.basename(process.cwd());
   const [defaultProjectName = fallbackProjectName] = (repoSlug && repoSlug.match(/[^/]+$/)) || [];
 
-  const responses = await inquirer.prompt([
-    {
+  const prompts = [];
```
patrickhulce, line 32: You'll want to add a typedef here, I'm not sure exactly what the type will be but it'll look something like this


```suggestion
  /** @type {Array<import('inquirer').WhateverThePromptTypeIs>} */
  const prompts = [];
```

packages/cli/src/wizard/wizard.js
```diff
@@ -29,33 +29,46 @@ async function runNewProjectWizard(options) {
...
       message: "What branch is considered the repo's trunk or main branch?",
       default: 'master',
-    },
-  ]);
+    });
+  }
+
+  const responses = {
+    ...options,
+    ...(await inquirer.prompt(prompts)),
```
patrickhulce, line 70: awesome! could we get a test case added? :) we should have some good examples of invoking the CLI already in there. maybe copy `healtcheck` to a new `wizard-no-prompts` one if you don't see a good existing place to add?

patrickhulce (changes requested):
thanks very much for the contribution @revolunet! great thinking!

packages/cli/src/wizard/wizard.js
revolunet: sorry, TS noob here, how to find the correct type ?

patrickhulce: if you're using vscode you can ctrl/cmd+click on the `import('inquirer')` bit and poke around the type definitions. In this case we want the same thing that `prompt` accepts, which looks like `Array<import('inquirer').Question>`

packages/cli/test/wizard.test.js
```diff
@@ -74,6 +74,34 @@ describe('wizard CLI', () => {
...
+          '--wizard=new-project',
+          '--basicAuth.username=admin',
+          '--basicAuth.password=io',
+          '--serverBaseUrl="https://lighthouse.hello.world"',
+          '--projectName="some-new-project"',
+          '--projectExternalUrl="https://somewhere.else"',
+          '--projectBaseBranch="master"',
+        ],
+        [
+          // '', // Just ENTER key to select "new-project"
```
patrickhulce, line 92: remove these?

packages/cli/test/wizard.test.js
```diff
@@ -74,6 +74,34 @@ describe('wizard CLI', () => {
...
+    it('should create a new project with arguments', async () => {
+      server.process.kill();
+      server = await startServer(server.sqlFile, ['--basicAuth.password=lighthouse']);
+
+      const {stdout, stderr, status} = await runWizardCLI(
+        [
+          '--wizard=new-project',
+          '--basicAuth.username=admin',
+          '--basicAuth.password=io',
+          '--serverBaseUrl="https://lighthouse.hello.world"',
```
patrickhulce, line 86: this is going to have to match the server you started

packages/cli/test/wizard.test.js
```diff
@@ -74,6 +74,34 @@ describe('wizard CLI', () => {
...
 
+    it('should create a new project with arguments', async () => {
+      server.process.kill();
+      server = await startServer(server.sqlFile, ['--basicAuth.password=lighthouse']);
+
+      const {stdout, stderr, status} = await runWizardCLI(
+        [
+          '--wizard=new-project',
+          '--basicAuth.username=admin',
+          '--basicAuth.password=io',
```
patrickhulce, line 85: same here though I don't think we need basicAuth as part of this test

patrickhulce (commented):
tried to give some pointers on the failing test :)

revolunet:
i cant figure out the issues on the typecheck


```
packages/cli/src/wizard/wizard.js:81:5 - error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

81     rootURL: responses.serverBaseUrl || options.serverBaseUrl,
       ~~~~~~~

  packages/utils/src/api-client.js:17:15
    17    * @param {{rootURL: string, fetch?: import('isomorphic-fetch'), URL?: typeof import('url').URL, extraHeaders?: Record<string, string>, basicAuth?: LHCI.ServerCommand.Options['basicAuth']}} options
                     ~~~~~~~
    The expected type comes from property 'rootURL' which is declared here on type '{ rootURL: string; fetch?: ((input: RequestInfo, init?: RequestInit | undefined) => Promise<Response>) | undefined; URL?: typeof URL | undefined; extraHeaders?: Record<...> | undefined; basicAuth?: { ...; } | undefined; }'

packages/cli/src/wizard/wizard.js:84:5 - error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

84     name: responses.projectName,
       ~~~~

  types/server.d.ts:20:9
    20         name: string;
               ~~~~
    The expected type comes from property 'name' which is declared here on type 'Pick<Project, "slug" | "createdAt" | "updatedAt" | "name" | "externalUrl" | "baseBranch">'

packages/cli/src/wizard/wizard.js:85:5 - error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

85     externalUrl: responses.projectExternalUrl,
       ~~~~~~~~~~~

  types/server.d.ts:21:9
    21         externalUrl: string;
               ~~~~~~~~~~~
    The expected type comes from property 'externalUrl' which is declared here on type 'Pick<Project, "slug" | "createdAt" | "updatedAt" | "name" | "externalUrl" | "baseBranch">'

packages/cli/src/wizard/wizard.js:86:5 - error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

86     baseBranch: responses.projectBaseBranch,
       ~~~~~~~~~~

  types/server.d.ts:23:9
    23         baseBranch: string;
               ~~~~~~~~~~
    The expected type comes from property 'baseBranch' which is declared here on type 'Pick<Project, "slug" | "createdAt" | "updatedAt" | "name" | "externalUrl" | "baseBranch">'

```

patrickhulce:
For the typecheck failures there are a couple options with our new dynamic sourcing but the most straightforward that should work for this situation is assert that they're defined after our prompt.

```js
if (!responses.serverBaseUrl) throw new Error('Failed to respond with a server URL');
if (!responses.projectName) throw new Error('Failed to respond with a project name');
```

If that doesn't work I've got a backup just so we can merge and I'll take a look at them later.

PR #541, opened 2021-01-29
feat(server) add ability to configure multiple cron tasks and specify which branches should be excluded or included during deletion.

MrFlashAccount:
Sometimes it would be extremely helpful to have some additional configuration options to describe logic, like:
I want to keep master builds for a year and all other branches for a short period of time

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -973,14 +973,23 @@ _Optional_ The "branch" on which to report the results. Defaults to the base bra
 #### `deleteOldBuildsCron`
 
 The configuration to automatically delete old records. This job only works when `storage.storageMethod` is set `sql`.
+Might be an array of objects, or just a single object.
 
-##### `deleteOldBuildsCron.schedule`
+##### `deleteOldBuildsCron[i].schedule`
 
 The [cron-style](https://crontab.guru/) schedule on which to collect results.
 
-##### `deleteOldBuildsCron.maxAgeInDays`
+##### `deleteOldBuildsCron[i].maxAgeInDays`
 
 A number of days to keep a records.
+
+##### `deleteOldBuildsCron[i].onlyBranches`
+
+Array of branch names, which should be deleted
+
+##### `deleteOldBuildsCron[i].filterOutBranches`
+
+Array of branch names, which should be excluded during cleanup
 
 #### `basicAuth`
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -987,7 +987,7 @@ A number of days to keep a records.
 
 Array of branch names, which should be deleted
 
-##### `deleteOldBuildsCron[i].filterOutBranches`
+##### `deleteOldBuildsCron[i].skipBranches`
 
 Array of branch names, which should be excluded during cleanup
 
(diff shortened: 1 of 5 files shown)


MrFlashAccount:
@googlebot I signed it!

docs/configuration.md
```diff
@@ -973,15 +973,24 @@ _Optional_ The "branch" on which to report the results. Defaults to the base bra
...
-##### `deleteOldBuildsCron.maxAgeInDays`
+##### `deleteOldBuildsCron[i].maxAgeInDays`
 
 A number of days to keep a records.
 
+##### `deleteOldBuildsCron[i].onlyBranches`
+
+Array of branch names, which should be deleted
+
+##### `deleteOldBuildsCron[i].filterOutBranches`
```
patrickhulce, line 990: WDYT about `skipBranches` to match the Lighthouse nomenclature of `onlyAudits`/`skipAudits`?

packages/server/src/cron/delete-old-builds.js
```diff
@@ -37,19 +49,42 @@ function startDeleteOldBuildsCron(storageMethod, options) {
...
+  const cronConfig = Array.isArray(options.deleteOldBuildsCron)
+    ? options.deleteOldBuildsCron
+    : [options.deleteOldBuildsCron];
+
+  cronConfig.forEach((config, index) => {
+    if (!config.schedule || !config.maxAgeInDays) {
+      throw new Error(
+        `Cannot configure schedule, because I haven't specified schedule field of maxAgeInDays field in item with index: ${index +
+          1}`
+      );
```
patrickhulce, line 71:
```suggestion
      throw new Error(`Invalid configuration for deleteOldBuilds[${index}]`);
```

WDYT about that?

patrickhulce (commented):
thanks for the contribution @MrFlashAccount! I think this is great 👍

docs/configuration.md
MrFlashAccount: This makes sense. Name corrected

packages/server/src/cron/delete-old-builds.js
MrFlashAccount: Sorry about that. Message fixed

packages/server/src/cron/delete-old-builds.js
```diff
@@ -37,19 +49,42 @@ function startDeleteOldBuildsCron(storageMethod, options) {
...
+   * @type {LHCI.ServerCommand.DeleteOldBuildsCron[]}
+   */
+  const cronConfig = Array.isArray(options.deleteOldBuildsCron)
+    ? options.deleteOldBuildsCron
+    : [options.deleteOldBuildsCron];
+
+  cronConfig.forEach((config, index) => {
+    if (!config.schedule || !config.maxAgeInDays) {
+      throw new Error(
+        `Can't configure schedule because you didn't specify 'schedule' field or 'maxAgeInDays' field in item with index: ${index +
```
patrickhulce, line 69: is there a middle ground between my suggestion and what you have here? :) 

I don't feel *very* strongly about my text, but definitely the context for `deleteOldBuilds` and using `index` instead of `index + 1` is important IMO.

patrickhulce: friendly ping, I'd love to get this merged for the imminent release

MrFlashAccount: Yea, I got what you mean. I didn't think about that 👍 I've replaced `index + 1` with `index`

MrFlashAccount:
@patrickhulce Maybe it's better to squash all these three commits? What do think?

patrickhulce (approved):

patrickhulce:
I squashed on merge, thanks again @MrFlashAccount! :)

PR #579, opened 2021-03-25
reference node 12 as supported version

dylangrandmont:
Users copy-pasting the existing starting template will run into issues such as https://github.com/GoogleChrome/lighthouse/issues/11953, since lighthouse CI will not work with node 10.


dylangrandmont:
@googlebot I signed it!

patrickhulce (approved):
great catch thanks @dylangrandmont!

PR #583, opened 2021-04-05
Update Node version in Github CI example

4vanger:
Update Node version to 12 to avoid bug described in https://github.com/GoogleChrome/lighthouse/issues/11854


patrickhulce:
thanks very much @4vanger this should be covered by #579 now 👍

PR #589, opened 2021-04-22
feat(server): enable passing of server host

PieterScheffers:
This PR adds the ability to pass a listening host to the Lighthouse CI server.
This makes it possible to let it bind to localhost or IPv4 with `0.0.0.0`.

**Background**
We want to run the docker image in our Kubernetes cluster. 
So we create a deployment with the `patrickhulce/lhci-server` image.
When accessing the server via the nginx ingress we keep getting a http 503 error.
After investigation the problem seems to be that the service is listening on IPv6 instead of IPv4.
So if we could pass it a IPv4 listenhost of `0.0.0.0` the server will use IPv4 instead of IPv6 and the problem will be solved.

Output netstat:
```
root@lhci-server-7fdf98f947-l8dwg:/usr/src/lhci# apt-get update && apt install -y net-tools
root@lhci-server-7fdf98f947-l8dwg:/usr/src/lhci# netstat -tulpn | grep LISTEN
tcp6       0      0 :::9001                 :::*                    LISTEN      25/node
```

### DIFF
--- a/packages/cli/src/server/server.js
+++ b/packages/cli/src/server/server.js
@@ -23,6 +23,9 @@ function buildCommand(yargs) {
       type: 'number',
       default: 9001,
     },
+    listenHost: {
+      type: 'string',
+    },
     'storage.storageMethod': {
       type: 'string',
       choices: ['sql', 'spanner'],
--- a/packages/cli/src/server/server.js
+++ b/packages/cli/src/server/server.js
@@ -23,7 +23,7 @@ function buildCommand(yargs) {
       type: 'number',
       default: 9001,
     },
-    listenHost: {
+    host: {
       type: 'string',
     },
     'storage.storageMethod': {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -887,6 +887,10 @@ Options:
 #### `port`
 
 The port for the server to listen on. A value of `0` will use a random available port.
+https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback
+#### `host`
+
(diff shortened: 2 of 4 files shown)


PieterScheffers:
@googlebot I signed it!

packages/cli/src/server/server.js
```diff
@@ -23,6 +23,9 @@ function buildCommand(yargs) {
       type: 'number',
       default: 9001,
     },
+    listenHost: {
```
patrickhulce, line 26: I can't think of what else we would use `host` for on the server, let's just use `host`?

patrickhulce (changes requested):
thanks @PieterScheffers this is a great addition! 

add a quick entry for `host` to our configuration docs and link to https://nodejs.org/api/net.html#net_server_listen_options_callback which discusses the default IPv6 behavior you ran into?

packages/cli/src/server/server.js
PieterScheffers: I first thought of using `host` but I thought it would bring some ambiguity with the `url` param. But I will change it to `host`.

PieterScheffers:
@patrickhulce Thanks for your feedback! I have added an entry in de configuration documentation about the `host` option.

packages/cli/src/server/server.js
patrickhulce: > I thought it would bring some ambiguity with the url param.

Oh like from the client-side `serverBaseUrl`? good thinking :) I'm not as worried about that because they're separate commands with their own namespaces, so I think we should be ok with `host` 👍

patrickhulce (approved):
thanks @PieterScheffers! this is great 💯

PR #595, opened 2021-04-28
fix(cli/upload): Fix GitHub status check

metalix2:
Related to #575 
Just getting it over the line (hopefully)

Took inspiration from the psi tests and set a mock server for github status check.

### DIFF
--- a/packages/cli/src/autorun/autorun.js
+++ b/packages/cli/src/autorun/autorun.js
@@ -1,5 +1,5 @@
 /**
- * @license Copyright 2019 Google Inc. All Rights Reserved.
+ * @license Copyright 2021 Google Inc. All Rights Reserved.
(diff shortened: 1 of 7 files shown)


packages/cli/src/autorun/autorun.js
```diff
@@ -1,5 +1,5 @@
 /**
- * @license Copyright 2019 Google Inc. All Rights Reserved.
+ * @license Copyright 2021 Google Inc. All Rights Reserved.
```
patrickhulce, line 2:
```suggestion
 * @license Copyright 2019 Google Inc. All Rights Reserved.
```

don't need to bump this one :)

patrickhulce (approved):
thanks @metalix2 and @raphaelbs for your help here!

PR #597, opened 2021-05-02
deps(docker): update node to  new lts 14

somehowchris:
This updates the docker base image to the "new" lts version of 14 which contains a bunch of vuln fixes with it (long live security)


patrickhulce:
Thanks @somehowchris! Is this working for you on your setup? I've yet to see LHCI work correctly on Node 14 due to issues in  SQL migration dependencies.

somehowchris:
Is this sql migrations thing happening on a specific SQL source? couldn't see any errors on sqlite

somehowchris:
Is there also a reason why not to move the image to alpine (docker found 146 system vulnerabilities)

patrickhulce:
> Is this sql migrations thing happening on a specific SQL source? couldn't see any errors on sqlite

The behavior that's typically observed is that the server just hangs without properly responding to requests, but no errors. But the server was responding properly?

> Is there also a reason why not to move the image to alpine (docker found 146 system vulnerabilities and with alpine 0)

Inertia, risk of change :) (last time someone changed the dockerfiles some dependencies were forgotten and broke some key pieces)

If the change were accompanied with some tests for the dockerfile so we'd finally have confidence in them, sounds great!

somehowchris:
> > Is this sql migrations thing happening on a specific SQL source? couldn't see any errors on sqlite
> 
> The behavior that's typically observed is that the server just hangs without properly responding to requests, but no errors. But the server was responding properly?

Ok, couldn't see any of them while creating a project/uploading data or viewing things.

> 
> > Is there also a reason why not to move the image to alpine (docker found 146 system vulnerabilities and with alpine 0)
> 
> Inertia, risk of change :) (last time someone changed the dockerfiles some dependencies were forgotten and broke some key pieces)
> 
> If the change were accompanied with some tests for the dockerfile so we'd finally have confidence in them, sounds great!

So how should these tests look like? checking for dependencies? How about actually moving the building and testing done with the github actions into docker? there are already client and server images

patrickhulce:
> So how should these tests look like?

Just pretty basic usage. I'm thinking of a script that builds the docker image and then attempts to use it while ensuring there are no errors. 

For client image, that means run the LHCI CLI in a test git repo and ensure it completes without failing (would have caught the fact that `git` was missing on our previous change).

For server image, that means startup the server, fill it with our test data, and make sure it can return it (would catch the node 14 bugs we've seen in the wild if they apply here).

I'll take a stab at this this week.

patrickhulce:
filed https://github.com/GoogleChrome/lighthouse-ci/issues/598 👍

patrickhulce:
@googlebot I consent.

patrickhulce (changes requested):
Docker tests already paying off! Looks like the v14 version lacks the dependencies necessary to use sqlite :)

patrickhulce:
FYI @somehowchris without movement I'm going to go ahead and close this one out, but please anyone feel free to come back to finish it up anytime. We'd love to accept it, thanks for getting involved! :)

somehowchris:
Hey @patrickhulce 

Due to an accident I am currently not able to work. Will be back in a few months (what doesn't kill me makes me stronger)

patrickhulce:
So sorry to hear that, hope you get better very soon!

PR #617, opened 2021-05-25
fix(cli): lhci open to process multiple urls

colourgarden:
### Description
- `lhci open` command currently silently fails if configuration `ci.collect.url` is an array of URLs (valid) rather than a string

### To reproduce
1. Set configuration:
  ```
  ci: {
    collect: {
      url: [
        'https://www.google.com/',
        'https://www.google.de/'
      ]
    }
  }
  ```
2. Run `lhci collect` to generate reports successfully
3. Run `lhci open`. No URL opened in browser (due to `continue`) but `Done!` message still printed to console

### DIFF
--- a/packages/cli/src/open/open.js
+++ b/packages/cli/src/open/open.js
@@ -39,7 +39,7 @@ async function runCommand(options) {
   }
 
   for (const lhr of representativeLhrs) {
-    if (options.url && (lhr.finalUrl !== options.url && !options.url.some(item => lhr.finalUrl.includes(item)))) continue;
+    if (targetUrls.size && !targetUrls.has(lhr.finalUrl)) continue;
 
     process.stdout.write(`Opening median report for ${lhr.finalUrl}...\n`);
     const tmpFile = tmp.fileSync({postfix: '.html'});
--- a/packages/cli/src/open/open.js
+++ b/packages/cli/src/open/open.js
@@ -37,6 +37,8 @@ async function runCommand(options) {
   if (!representativeLhrs.length) {
     process.stdout.write('No available reports to open. ');
   }
+  
+  const targetUrls = new Set([].concat(options.url || []));
 
   for (const lhr of representativeLhrs) {
     if (targetUrls.size && !targetUrls.has(lhr.finalUrl)) continue;
--- a/packages/cli/src/open/open.js
+++ b/packages/cli/src/open/open.js
@@ -43,7 +43,7 @@ async function runCommand(options) {
     (options.url || []);
 
   for (const lhr of representativeLhrs) {
-    if (targetUrls.size && !targetUrls.has(lhr.finalUrl)) continue;
+    if (targetUrls.length && !targetUrls.includes(lhr.finalUrl)) continue;
 
     process.stdout.write(`Opening median report for ${lhr.finalUrl}...\n`);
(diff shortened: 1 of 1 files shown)


colourgarden:
@googlebot I signed it!

packages/cli/src/open/open.js
```diff
@@ -39,7 +39,7 @@ async function runCommand(options) {
   }
 
   for (const lhr of representativeLhrs) {
-    if (options.url && lhr.finalUrl !== options.url) continue;
+    if (options.url && (lhr.finalUrl !== options.url && !options.url.some(item => lhr.finalUrl.includes(item)))) continue;
```
patrickhulce, line 42: how about a target URLs set? I think this will fail when url is a string, not an array

`const targetUrls = new Set([].concat(options.url || []));`

```suggestion
    if (targetUrls.size && !targetUrls.has(lhr.finalUrl)) continue;
```

patrickhulce (changes requested):
thanks very much @colourgarden!

this is mostly an accident for a different reason. the config isn't supposed to apply to the `open` command, but it seems reasonable to support anyhow.

packages/cli/src/open/open.js
colourgarden: @patrickhulce Thanks for the swift feedback! I did actually test my solution with both string and array and it does work with both. However, I'd defer to you as creator/maintainer on which approach is best. I've quickly tested your solution locally and it also works as expected 👍 

I assume `targetUrls` is best set on line 40 after the `representativeLhrs` check?

patrickhulce: > it does work with both

Only when the `finalUrl` of all the reports matches `options.url` ;) `TypeError: options.url.some is not a function` if you have an LHR with a non-matching URL that should be skipped and `options.url` is a string

> I assume targetUrls is best set on line 40 after the representativeLhrs check?

Yep SGTM, thanks! 👍

patrickhulce (approved):
thanks @colourgarden!

packages/cli/src/open/open.js
```diff
@@ -37,9 +37,11 @@ async function runCommand(options) {
   if (!representativeLhrs.length) {
     process.stdout.write('No available reports to open. ');
   }
+  
+  const targetUrls = new Set([].concat(options.url || []));
```
patrickhulce, line 41: aw shoot, sorry @colourgarden but typescript doesn't like my too clever suggestion 😅 
```suggestion
  const targetUrls = typeof options.url === 'string' ? 
    [options.url] : 
    (options.url || []);
```

packages/cli/src/open/open.js
```diff
@@ -37,9 +37,11 @@ async function runCommand(options) {
...
 
   for (const lhr of representativeLhrs) {
-    if (options.url && lhr.finalUrl !== options.url) continue;
+    if (targetUrls.size && !targetUrls.has(lhr.finalUrl)) continue;
```
patrickhulce, line 44:
```suggestion
    if (targetUrls.length && !targetUrls.includes(lhr.finalUrl)) continue;
```

patrickhulce (commented):
sorry for steering you in the wrong direction!

patrickhulce (approved):

patrickhulce:
@googlebot I consent.

PR #631, opened 2021-06-16
feat(cli): add psi strategy in option for psi runner, keeping mobile …

loickaczmarek:
Hi, i need to use mobile and desktop strategy for psi.

So, i add option for lhci collect.

use:

lhci collect --psiStrategy='mobile'
lhci collect --psiStrategy='desktop'

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -218,6 +218,7 @@ Options:
   --puppeteerLaunchOptions   The object of puppeteer launch options
   --psiApiKey                [psi only] The API key to use for PageSpeed Insights runner method.
                              You do not need to use this unless you wrote a custom version.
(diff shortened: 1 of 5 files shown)


patrickhulce (changes requested):
thanks @loickaczmarek!

could you add this to the yargs command and documentation as well?

https://github.com/GoogleChrome/lighthouse-ci/blob/ecbde706456eca05b6d0e78ce6659a773771de94/packages/cli/src/collect/collect.js#L46-L48

https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md#psiapiendpoint

patrickhulce (approved):
thanks so much! 🎉

PR #633, opened 2021-06-18
update cli version to 0.8 in docs

panmau:
While doing the first time set-up I noticed that the getting started docs are not up-to date with the current CLI version so I updated all `npm install` commands to 0.8.

Also I noticed that the migration guide doesn't include guidance on 0.7 to 0.8 yet. Maybe you can add those?

### DIFF
--- a/docs/complex-setup.md
+++ b/docs/complex-setup.md
@@ -24,7 +24,7 @@ fi
 
 npm run deploy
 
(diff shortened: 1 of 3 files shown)


panmau:
@googlebot I signed it!

patrickhulce (approved):
thanks very much @panmau!

PR #636, opened 2021-06-20
Localization for closed networks

EhhChris:
As discussed in https://github.com/GoogleChrome/lighthouse-ci/discussions/635 - this provides two relatively simple changes:

- Makes use of fontsource to pull in roboto, roboto-mono, and material icons instead of using the googleapis font CDN (without use of a proxy these lead to rendering issues on the server-ui and viewer)
- An optional use of $VIEWER_ORIGIN environment variable to configure a self-hosted [viewer](https://github.com/GoogleChrome/lighthouse/tree/master/lighthouse-viewer) instead of linking to the public internet copy on github

As I mentioned in the discussion, I have relatively little UI experience and if there's a better way to accomplish these localizations I'm open to learning other options. I think the alternative I was finding a lot of while googling was using something like webpack to produce the public.html for the font scripts if we wanted this to be optional.

Finally, I think both of these changes were relatively trivial but if a test is wanted I could look into learning some jest. Thanks!

### DIFF
--- a/docs/complex-setup.md
+++ b/docs/complex-setup.md
@@ -24,7 +24,7 @@ fi
 
 npm run deploy
 
-npm install -g @lhci/cli@0.7.x
+npm install -g @lhci/cli@0.8.x
 lhci healthcheck --fatal
 lhci collect --url=http://localhost:9000/index.html
 lhci assert --preset="lighthouse:recommended"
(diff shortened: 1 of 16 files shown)


EhhChris:
@googlebot I signed it!

EhhChris:
Will re-open with a fixed (clean) history to make the bot happy. Configuration issue on my end locally that probably should have been fixed anyway.

Issue #461, opened 2020-09-29
LHCI server open HTML report to external url 


patriciamarchand:
The LHCI Dashboard open the report HTML to an external URL (https://googlechrome.github.io/lighthouse/viewer/). 
Does it have a configuration parameter to open the report but not on an external URL. I would like to avoid that. 

![image](https://user-images.githubusercontent.com/6225238/94612889-bd92b300-0271-11eb-80f0-362793a93627.png)

Thanks for this great tool!

patrickhulce:
Thanks for filing @patriciamarchand! There is no such parameter, but we can let this issue track building one :)

mfjerome:
Question: Why is the lighthouse report viewer not included in the lighthouse report server? Before spinning it up, I assumed that the goal of having a report server was to host the audit reports so that the developers can view and compare them between commits, without having to rely on the viewer hosted in github. Did I mimsunderstood something?

Would it be feasible to do so, and what would be the best way to accomplish it? Maybe I can help.

patrickhulce:
> Why is the lighthouse report viewer not included in the lighthouse report server?

The code for the viewer is in the upstream Lighthouse core repo and not easily hosted anywhere else due to the Firebase Auth and GitHub Gist features it offers. It sounds like most people don't care about those features of the viewer and would prefer to keep everything on their server even if its just the static report, so that's what this issue tracks.

> I assumed that the goal of having a report server was to host the audit reports so that the developers can view and compare them between commits, without having to rely on the viewer hosted in github

The goal of the server is to store reports long term and compare between commits. Delegating to existing infrastructure that enables viewing individual reports fit well within that goal with eventual support for providing more value on top of the historical reports. 

> Would it be feasible to do so

Yes, that's what this issue is tracking. We just need to add a route to the server that grabs an LHR from storage and renders it inside the HTML template. Fairly straightforward with our existing utils around this if you'd like to take a crack it :)

patriciamarchand:
For my understanding, I had a look in the DB. I found a JSON in the table "run" under the column "lhr". Are you using this data to generate the diff report?

For the Lighthouse report viewer whitch data are you using? I did not find the HTML code in the BD. Are you using the data for the table "statistics"? The command lhci autorun generate JSON and HTML file. Are you storing both?

Thanks

patrickhulce:
> Are you using this data to generate the diff report?

Yes

> For the Lighthouse report viewer whitch data are you using?

The same JSON data.

[...]

PR #637, opened 2021-06-20
Updates to support closed network

EhhChris:
As discussed in https://github.com/GoogleChrome/lighthouse-ci/discussions/635 - this provides two relatively simple changes:

- Makes use of fontsource to pull in roboto, roboto-mono, and material icons instead of using the googleapis font CDN (without use of a proxy these lead to rendering issues on the server-ui and viewer)
- An optional use of $VIEWER_ORIGIN environment variable to configure a self-hosted [viewer](https://github.com/GoogleChrome/lighthouse/tree/master/lighthouse-viewer) instead of linking to the public internet copy on github

As I mentioned in the discussion, I have relatively little UI experience and if there's a better way to accomplish these localizations I'm open to learning other options. I think the alternative I was finding a lot of while googling was using something like webpack to produce the public.html for the font scripts if we wanted this to be optional.

Finally, I think both of these changes were relatively trivial but if a test is wanted I could look into learning some jest. Thanks!

(re-created PR with fixed git history for the bot)

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -20,8 +20,6 @@
     "start:storybook": "start-storybook -p 6006 --no-dll"
   },
   "dependencies": {
-    "@fontsource/material-icons": "^4.4.5",
-    "@fontsource/roboto": "^4.4.5",
     "@lhci/utils": "0.1.0",
     "bluebird": "^3.7.2",
     "body-parser": "^1.18.3",
@@ -43,7 +41,10 @@
     "plotly.js": "^1.48.3",
     "preact": "^10.0.1",
     "preact-async-route": "^2.2.1",
-    "preact-router": "^3.1.0"
+    "preact-router": "^3.1.0",
+    "@fontsource/material-icons": "^4.4.5",
+    "@fontsource/roboto": "^4.4.5",
+    "@fontsource/roboto-mono": "^4.4.5"
   },
   "alias": {
     "isomorphic-fetch": "clsx"
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -20,6 +20,8 @@
     "start:storybook": "start-storybook -p 6006 --no-dll"
   },
   "dependencies": {
+    "@fontsource/material-icons": "^4.4.5",
+    "@fontsource/roboto": "^4.4.5",
     "@lhci/utils": "0.1.0",
     "bluebird": "^3.7.2",
     "body-parser": "^1.18.3",
--- a/packages/server/src/ui/components/lhr-viewer-link.jsx
+++ b/packages/server/src/ui/components/lhr-viewer-link.jsx
@@ -13,7 +13,8 @@ import './lhr-viewer-link.css';
 
 /** @param {LH.Result} lhr */
 export function openLhrInClassicViewer(lhr) {
-  const VIEWER_ORIGIN = 'https://googlechrome.github.io';
+  // Allows for self-hosted viewer uri instead of linking externally
+  const VIEWER_ORIGIN = process.env.VIEWER_ORIGIN != null ? process.env.VIEWER_ORIGIN : 'https://googlechrome.github.io';
   // Chrome doesn't allow us to immediately postMessage to a popup right
   // after it's created. Normally, we could also listen for the popup window's
   // load event, however it is cross-domain and won't fire. Instead, listen
--- a/packages/server/src/ui/components/lhr-viewer-link.jsx
+++ b/packages/server/src/ui/components/lhr-viewer-link.jsx
@@ -14,7 +14,7 @@ import './lhr-viewer-link.css';
 /** @param {LH.Result} lhr */
 export function openLhrInClassicViewer(lhr) {
   // Allows for self-hosted viewer uri instead of linking externally
-  const VIEWER_ORIGIN = process.env.VIEWER_ORIGIN != null ? process.env.VIEWER_ORIGIN : 'https://googlechrome.github.io';
+  const VIEWER_ORIGIN = process.env.VIEWER_ORIGIN || 'https://googlechrome.github.io';
   // Chrome doesn't allow us to immediately postMessage to a popup right
   // after it's created. Normally, we could also listen for the popup window's
   // load event, however it is cross-domain and won't fire. Instead, listen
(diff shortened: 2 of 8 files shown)


packages/server/src/ui/entry.jsx
```diff
@@ -7,6 +7,11 @@
 import {h, render} from 'preact';
 import {App} from './app.jsx';
 
+// Fontsource for fonts/icons instead of google cdn
+import "@fontsource/roboto/400.css";
```
patrickhulce, line 11: missing roboto mono?

packages/server/package.json
```diff
@@ -20,6 +20,8 @@
     "start:storybook": "start-storybook -p 6006 --no-dll"
   },
   "dependencies": {
+    "@fontsource/material-icons": "^4.4.5",
+    "@fontsource/roboto": "^4.4.5",
```
patrickhulce, line 24: I think these can be devDependencies? the frontend source is *built* and published with the npm package

packages/server/src/ui/components/lhr-viewer-link.jsx
```diff
@@ -13,7 +13,8 @@ import './lhr-viewer-link.css';
 
 /** @param {LH.Result} lhr */
 export function openLhrInClassicViewer(lhr) {
-  const VIEWER_ORIGIN = 'https://googlechrome.github.io';
+  // Allows for self-hosted viewer uri instead of linking externally
+  const VIEWER_ORIGIN = process.env.VIEWER_ORIGIN != null ? process.env.VIEWER_ORIGIN : 'https://googlechrome.github.io';
```
patrickhulce, line 17: I appreciate the narrowness, but if `VIEWER_ORIGIN` is empty string we'd probably still want to use the default :)

```suggestion
  const VIEWER_ORIGIN = process.env.VIEWER_ORIGIN || 'https://googlechrome.github.io';
```

this will only work though if `VIEWER_ORIGIN` is defined when the project is *built* so the npm package won't work out of the box. Should it accept a URL flag/`localStorage`/`Cookie` as well?

patrickhulce (commented):
thanks @EhhChris!

cactysman:
> An optional use of $VIEWER_ORIGIN environment variable to configure a self-hosted viewer

Great move, I just stumbled across this myself!
Even though I trust Google with this data, having a way to use a custom viewer is always a good thing.

packages/server/src/ui/components/lhr-viewer-link.jsx
cactysman: > this will only work though if `VIEWER_ORIGIN` is defined when the project is _built_

Couldn't the server tell the frontend about the viewer URL somehow? Maybe a configurable setting per project even with whatever's set for the server via env as default?

patrickhulce: > Couldn't the server tell the frontend about the viewer URL somehow?

Yes, as suggested, cookies would be one way of doing that, but a more heavy-handed explicit configuration per project could also work (way more work on the UI-side though to fetch the project settings and thread through all the layers that require a report render button). I don't suspect that this is terribly important to configure *per project* either. It's either a closed instance of the server or it's not in most cases.

cactysman: > I don't suspect that this is terribly important to configure *per project* either. It's either a closed instance of the server or it's not in most cases.

True, that might have been an overzealous thought 🙂

packages/server/package.json
EhhChris: Yep! Moved them over in both viewer/server.

packages/server/src/ui/entry.jsx
EhhChris: Somehow I convinced myself that it wasn't included in the server public.html, but you're right- fixed.

packages/server/src/ui/components/lhr-viewer-link.jsx
EhhChris: So, having looked around previously- and not having considered that the environment variable would be not useful as part of the npm package distribution because the UI is built already... maybe it makes more sense to solve this particular piece with #461 instead? If I understand that issue correctly it would pull a component of the lighthouse package into this one, minus the firebase auth.

patrickhulce: > maybe it makes more sense to solve this particular piece with #461 instead?

It definitely does make more sense, but making this configurable still seems fine. Let's leave it as-is and plan to tackle #461 👍

EhhChris: Sounds like a plan, I'm happy to take a crack at it this weekend. Thanks for all the help!

patrickhulce: Awesome that'd be great! It shouldn't be *too* bad, I'm thinking a single route on the server that accepts a runId and then returns the HTML report as-is from 

https://github.com/GoogleChrome/lighthouse-ci/blob/438f3e1c60ce1b4396024fad01220e7a6e376246/packages/utils/src/saved-reports.js#L53

The harder part will be replacing the existing buttons that link to the viewer to use the new route with `runId` instead.

patrickhulce (approved):
LGTM! 

let's see what CI has to say 😄

patrickhulce:
> let's see what CI has to say 😄

looks like we have some lint nitpicks around `"` v. `'`

EhhChris:
> > let's see what CI has to say 😄
> 
> looks like we have some lint nitpicks around `"` v. `'`

Woops! Got those sorted away.

packages/server/src/ui/components/lhr-viewer-link.jsx
EhhChris: > Awesome that'd be great! It shouldn't be _too_ bad, I'm thinking a single route on the server that accepts a runId and then returns the HTML report as-is from

Wanted to report back that I did end up taking a stab at this over the weekend and got to a point where I _think_ what I have will work, but kept running into issues with the seed data the server generates. It looks like the seed data is based on a pre-release version lighthouse 6 and is no longer compatible with the render code (can be seen with the current external lighthouse viewer from a seed data generated build).

When I find some more time this week I'll go through with adding a test project and generating some lighthouse 8 reports to validate it works and cleaning up the buttons for a PR.

OhArthits (approved):

PR #650, opened 2021-07-07
feat: support cjs config file extension

remcohaszing:
This allows users to use Lighthouse CI in a project which has `"type": "module"` in their `package.json` file.

I also looked into supporting ESM configuration files, but since `import()` is asynchronous, this is a lot more work than just supporting `.cjs` files in a code base I’m not very familiar with.

### DIFF
--- a/packages/utils/src/lighthouserc.js
+++ b/packages/utils/src/lighthouserc.js
@@ -11,6 +11,8 @@ const yaml = require('js-yaml');
 const _ = require('./lodash.js');
 
 const RC_FILE_NAMES = [
+  '.lighthouserc.cjs',
+  'lighthouserc.cjs',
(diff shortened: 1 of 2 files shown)


patrickhulce (commented):
thanks for filing @remcohaszing! could we add a test case for this or is it prohibitively difficult to setup?

remcohaszing:
> thanks for filing @remcohaszing! could we add a test case for this or is it prohibitively difficult to setup?

Done. That was actually pretty straight forward. :)

patrickhulce (approved):
thanks @remcohaszing!

fsmaia:
Amazing!!
It just is missing an update at `docs/getting-started.md`.

Issue #652, opened 2021-07-14
Add option to configure migrations table name


tim-yao:
Currently LHCI server will create a table named `SequelizeMeta` in the db. However, our server db only support lowercase table name. Which cause a problem to use it.

According to the docs from `sequelize` https://sequelize.org/v4/manual/tutorial/migrations.html#migration-storage, I should able to use a different table name by below config:
```
    // Use a different table name. Default: SequelizeMeta
    "migrationStorageTableName": "sequelize_meta"
```
See another suggestion from https://stackoverflow.com/a/56692455/1212791

I see LHCI support passing options to `sequelize` in docs configuration:
https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md#storagesequelizeoptions

I tried below in my server, but it doesn't change anything, the db table created is still `SequelizeMeta`. If anyone can tell what I suppose to do, that will be great! Thanks!
```
createServer({
  port: 3000,
  storage: {
    storageMethod: 'sql',
    sqlDialect: 'mysql',
    sqlConnectionUrl: dbUrl,
    sequelizeOptions: {
      migrationStorage: 'sequelize',
      migrationStorageTableName: 'sequelize_meta'
    }
  },
 ...
```

patrickhulce:
Thanks for filing @tim-yao! LHCI initializes the migration runner separately so sequelizeOptions for migrations won't take effect.

https://github.com/GoogleChrome/lighthouse-ci/blob/31e731a338ee50d241ea9ae40400748bd3901c4b/packages/server/src/api/storage/sql/sql.js#L136-L144

`storageOptions: {tableName: 'sequelize_meta'}` is what you'd need here. PR welcome for a `sqlMigrationOptions: {tableName?: string}` option to be passed through :)

tim-yao:
Thanks @patrickhulce! I patched the `sql.js` to add `storageOptions: {tableName: 'sequelize_meta'}` and it works. I will have a look at how to add the option in server and submit a PR if I can make it.

PR #653, opened 2021-07-16
feat: add support for sequlize migration options

tim-yao:
For suggestion from https://github.com/GoogleChrome/lighthouse-ci/issues/652


tim-yao:
@googlebot I signed it!

packages/server/test/mysql-server-migration.test.js
```diff
@@ -0,0 +1,47 @@
...
+        sqlMigrationOptions: {tableName: 'sequelize_meta'},
+      },
+    });
+
+    state.port = port;
+    state.closeServer = close;
+    state.storageMethod = storageMethod;
+  });
+
+  afterAll(() => {
```
patrickhulce, line 42: could you add one extra `it('should use the renamed table'` to manually query to DB and make sure it uses the `sequelize_meta` version? I think all of these tests would pass without implementing `sqlMigrationOptions` support (it would just be ignored)

patrickhulce (commented):
thanks very much for the contribution @tim-yao! :D

impl LGTM, just a test request

packages/server/test/mysql-server-migration.test.js
tim-yao: Thanks @patrickhulce I have added the test but not sure if it's the best practice, please let me know if anything need to be improved. 🤓

packages/server/test/mysql-server-migration.test.js
```diff
@@ -0,0 +1,59 @@
...
+    state.closeServer();
+  });
+
+  describe('sqlMigrationOptions', () => {
+    it('should use the renamed table', async () => {
+      const sequelize = new Sequelize(process.env.MYSQL_DB_URL);
+      const tables = await sequelize.query('show tables');
+      await sequelize.close();
```
patrickhulce, line 53: could you put this line in a `finally` block? if our query fails we still want the server connection to close so test process will exit :)

patrickhulce (approved):

patrickhulce:
also fyi have some lint errors to fix :) https://github.com/GoogleChrome/lighthouse-ci/pull/653/checks?check_run_id=3094377453#step:9:33

packages/server/test/mysql-server-migration.test.js
tim-yao: Thanks @patrickhulce! Updated.

patrickhulce:
@tim-yao I think we're running into issues with the test being run in the same set as the original mysql test because they're using the same DB. let's just go ahead and *change* the existing test to use the custom migration table name and add our assertion there rather than re-run the mysql set. WDYT?

tim-yao:
> @tim-yao I think we're running into issues with the test being run in the same set as the original mysql test because they're using the same DB. let's just go ahead and _change_ the existing test to use the custom migration table name and add our assertion there rather than re-run the mysql set. WDYT?

That make sense to me. Will do it.

patrickhulce:
> any idea? I get these errors in test with the new migration configs.I can see them in the CI as well.

You haven't pushed the latest version with the suggestion so I'm not sure. Those errors are expected with the version of the code that has been pushed here (the databases are being reused between 2 tests with unexpected state).

PR #659, opened 2021-07-24
(chore: version) - Updated the package.json to the version which is published in upstream.

riginoommen:


hixio-mh (approved):

patrickhulce (commented):
Thanks @riginoommen! Versions are tracked via git tags and updated at publish time, though not in these files themselves.

PR #677, opened 2021-08-19
fix(utils): throw on non-200 version response

### LINKED ISSUE #676: helthCheck of LHCI server reachable might be incorrect
**Describe the bug**

This bug exist in `@lhci/cli`.

helthCheck of `LHCI server reachable` might be incorrect.

[the code](https://github.com/GoogleChrome/lighthouse-ci/blob/72107f3bf4/packages/cli/src/healthcheck/healthcheck.js#L101) only check the the lenth of response, [getVersion](https://github.com/GoogleChrome/lighthouse-ci/blob/72107f3bf462ab60596f576967ff1a5e0aad622b/packages/utils/src/api-client.js#L165)  didn't not check the response status. If the lighthouse server is protected by firewall, response status may be 403 so it is unreachable actually. 

**To Reproduce**

1. try to response 403 status in `/version` path
2. helthCheck of `LHCI server reachable` still pass

**Expected behavior**

`LHCI server reachable` only pass when the `/version` response status is 200.

**Logs/Screenshots**

**Additional context**

patrickhulce:
fixes #676

### DIFF
--- a/packages/utils/src/api-client.js
+++ b/packages/utils/src/api-client.js
@@ -170,6 +170,7 @@ class ApiClient {
     const body = response.text();
     
     if (!response.ok) {
+      /** @type {Error & {status: number, body: any}} */
       const error = new Error(`Unexpected status code ${response.status}\n  ${body}`);
       error.status = response.status;
       error.body = body;
--- a/packages/utils/src/api-client.js
+++ b/packages/utils/src/api-client.js
@@ -166,7 +166,17 @@ class ApiClient {
     const response = await this._fetch(this._normalizeURL('/version').href, {
       headers: {...this._extraHeaders},
     });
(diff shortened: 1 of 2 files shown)


packages/utils/src/api-client.js
```diff
@@ -166,7 +166,18 @@ class ApiClient {
     const response = await this._fetch(this._normalizeURL('/version').href, {
       headers: {...this._extraHeaders},
     });
-    return response.text();
+    
+    const body = response.text();
+    
+    if (!response.ok) {
+      /** @type {Error & {status: number, body: any}} */
```
patrickhulce, line 173:
```suggestion
      /** @type {Error & {status?: number, body?: any}} */
```

PR #678, opened 2021-08-21
Decode URL's for display

ntkoopman:
Some of the URLs I'm using are in Chinese, which currently display urlencoded and are therefore hard to differentiate. This just decodes them for display (similar to the address bar).


patrickhulce (approved):
thanks very much @ntkoopman!

PR #679, opened 2021-08-23
Specify node.js version in Github Actions example

j0nas:
Specify node.js version for `actions/setup-node@v1` step in example Github Actions code. Without this, the snippet [currently fails](https://gist.github.com/j0nas/06fc18b10d74d9b8ca3ad8f0db9cac74) as it defaults to an earlier version of node.js which raises a SyntaxError for the current version of the module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
+        with:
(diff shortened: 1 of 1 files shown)


j0nas:
@googlebot I signed it!

patrickhulce (approved):
thanks @j0nas!

PR #688, opened 2021-09-11
fixes #687: ensure history commit message easy to read

### LINKED ISSUE #687: compare screen commit messages difficult to read
**Describe the bug**
When we go to the compare screen and click on the compare button we see that git commits. When the commit message is to long we see and the commit text flows over other commits

**To Reproduce**
1. Go to a project
2. Click on a commit. now your are on the compare screen
3. Click on the compare button 
4. When text is to long of the commits it flows over.

**Expected behavior**
The text doesn't flow over

**Logs/Screenshots**
chrome_kGUQ2K50yl

 - OS: Windows 10
 - Browser chrome 
 - Version [e.g. 93]

Additional context:
This overflow is caused by the whitespace css property on the .build-hash-selector__commit selector. When I disable the whitspace: pre property the issue is resolved see screenshot below

image

geoffreysamper:


packages/server/src/ui/routes/build-view/build-hash-selector.css
```diff
@@ -122,7 +121,6 @@
   flex-grow: 1;
   flex-shrink: 10;
   text-overflow: ellipsis;
-  white-space: pre;
```
patrickhulce, line 125: instead of deleting this line, let's just use `nowrap`

patrickhulce (changes requested):

patrickhulce:
thanks very much @geoffreysamper !

packages/server/src/ui/routes/build-view/build-hash-selector.css
geoffreysamper: @patrickhulce tested it with nowrap
image
tested it with wrap
image

I think the wrap is better but it your call

patrickhulce: `nowrap` was the intended design here, let's stick with that

geoffreysamper: I changed it to nowrap

patrickhulce (approved):
awesome thank you!

PR #696, opened 2021-09-24
Allow for further configuration of psiCollectCron

### LINKED ISSUE #695: Additional psi configurations
**Is your feature request related to a problem? Please describe.**

1. I'm attempting to run the psi cron job against all of our public pages on prod. We have a high number of them (~40), so sending all of the fetch requests in parallel overwhelms the pagespeed insights api and tanks out with the error `PSI collection failure for Production LHCI Cron Job - Home: PSI Failed (500): Lighthouse returned error: Something went wrong.` (I've been able to replicate this outside of this package). I'd like to be able to limit the number of parallel requests sent to the pagespeed insights api.
2. Also, a few of our pages error out when requesting the `pwa` category. I'd like to be able to opt out of categories from the config.

The above could be 2 separate issues, but I've grouped them together as these could both be optional psi configurations. While we're here, we could make other details configurable (see #420)

**Describe the solution you'd like**
I think the config could look something like:
```.js
psiCollectCron: {
    ...currentPsiCollectCronConfig,
    sites: [
        {
            ...currentSiteConfig,
            maxNumberOfParallelUrls: number, // default: Infinity (run all urls in parallel - current behavior)
            excludeCategories: ['performance' | 'accessibility' | 'best-practices' | 'pwa' | 'seo'], // default: [] (run all categories - current behavior)
            strategy: 'desktop' | 'mobile' // default: 'mobile' (current behavior)
        },
        ...etc
    ]
}
```

It wouldn't be difficult to set up an async counter to ensure a max number of requests are being processed in parallel. Even easier to just pass config values to each run.

**Describe alternatives you've considered**
Without being able to configure the number of parallel calls ([hardcoded here](https://github.com/GoogleChrome/lighthouse-ci/blob/f5c5ba5e9b507872fce2d7512bc7e5701bb9e6ce/packages/server/src/cron/psi-collect.js#L44-L63)) or being able to configure the categories ([hardcoded here](https://github.com/GoogleChrome/lighthouse-ci/blob/f5c5ba5e9b507872fce2d7512bc7e5701bb9e6ce/packages/utils/src/psi-client.js#L29-L56)), I don't see alternatives aside from setting up our own implementation of this. Not that this would be terribly difficult, but I'd rather make these few pieces configurable since it would be _almost_ identical to this package.

**Additional context**
Happy to open a PR, just wanted to get a pulse on if there is any opposition and why.

jonathantredway:
closes #695

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1002,7 +1002,7 @@ _Optional_ The "branch" on which to report the results. Defaults to the base bra
 
 ##### `psiCollectCron.sites[i].maxNumberOfParallelUrls`
 
-_Optional_ The maximum number of requests to send to the pagespeed insights api concurrently. Defaults to `Infinity` (all urls sent in parallel).
+_Optional_ The maximum number of requests to send to the PageSpeed Insights API concurrently. Defaults to `Infinity` (all urls sent in parallel).
 
 ##### `psiCollectCron.sites[i].categories`
 
@@ -1010,7 +1010,7 @@ _Optional_ An array containing the categories to test for each url in this site.
 
 ##### `psiCollectCron.sites[i].strategy`
 
-_Optional_ The strategy that the pagespeed insights api should use when testing each url in this site. Can be either `desktop` or `mobile`. Defaults to `mobile`.
+_Optional_ The strategy that the PageSpeed Insights API should use when testing each url in this site. Can be either `desktop` or `mobile`. Defaults to `mobile`.
 
 #### `deleteOldBuildsCron`
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1000,6 +1000,18 @@ _Optional_ The human friendly label for this set of URLs to use when logging sta
 
 _Optional_ The "branch" on which to report the results. Defaults to the base branch of the project referenced by `projectSlug`.
 
+##### `psiCollectCron.sites[i].maxNumberOfParallelUrls`
+
+_Optional_ The maximum number of requests to send to the pagespeed insights api concurrently. Defaults to `Infinity` (all urls sent in parallel).
+
+##### `psiCollectCron.sites[i].categories`
+
+_Optional_ An array containing the categories to test for each url in this site. Defaults to `['performance', 'accessibility', 'best-practices', 'pwa', 'seo']` (all categories).
+
+##### `psiCollectCron.sites[i].strategy`
+
+_Optional_ The strategy that the pagespeed insights api should use when testing each url in this site. Can be either `desktop` or `mobile`. Defaults to `mobile`.
+
 #### `deleteOldBuildsCron`
 
 The configuration to automatically delete old records. This job only works when `storage.storageMethod` is set `sql`.
--- a/packages/server/src/cron/psi-collect.js
+++ b/packages/server/src/cron/psi-collect.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const {CronJob} = require('cron');
+const Bluebird = require('bluebird');
 const {getGravatarUrlFromEmail} = require('@lhci/utils/src/build-context');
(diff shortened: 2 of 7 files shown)


packages/server/src/cron/psi-collect.js
```diff
@@ -17,10 +18,20 @@ const {normalizeCronSchedule} = require('./utils');
  * @return {Promise<void>}
  */
 async function psiCollectForProject(storageMethod, psi, site) {
-  const {urls, projectSlug, numberOfRuns = 5} = site;
+  const {
+    urls,
+    projectSlug,
+    numberOfRuns = 5,
+    maxNumberOfParallelUrls = Infinity,
+    categories = ['performance', 'accessibility', 'best-practices', 'pwa', 'seo'],
```
patrickhulce, line 26: maybe don't default it here, so it can fallback to the default in psi-client?

docs/configuration.md
```diff
@@ -1000,6 +1000,18 @@ _Optional_ The human friendly label for this set of URLs to use when logging sta
 
 _Optional_ The "branch" on which to report the results. Defaults to the base branch of the project referenced by `projectSlug`.
 
+##### `psiCollectCron.sites[i].maxNumberOfParallelUrls`
+
+_Optional_ The maximum number of requests to send to the pagespeed insights api concurrently. Defaults to `Infinity` (all urls sent in parallel).
```
patrickhulce, line 1005:
```suggestion
_Optional_ The maximum number of requests to send to the PageSpeed Insights API concurrently. Defaults to `Infinity` (all urls sent in parallel).
```

docs/configuration.md
```diff
@@ -1000,6 +1000,18 @@ _Optional_ The human friendly label for this set of URLs to use when logging sta
...
+
+##### `psiCollectCron.sites[i].categories`
+
+_Optional_ An array containing the categories to test for each url in this site. Defaults to `['performance', 'accessibility', 'best-practices', 'pwa', 'seo']` (all categories).
+
+##### `psiCollectCron.sites[i].strategy`
+
+_Optional_ The strategy that the pagespeed insights api should use when testing each url in this site. Can be either `desktop` or `mobile`. Defaults to `mobile`.
```
patrickhulce, line 1013:
```suggestion
_Optional_ The strategy that the PageSpeed Insights API should use when testing each url in this site. Can be either `desktop` or `mobile`. Defaults to `mobile`.
```

packages/utils/src/psi-runner.js
```diff
@@ -19,20 +19,22 @@ class PsiRunner {
 
   /**
    * @param {string} url
-   * @param {{psiApiKey?: string, psiApiEndpoint?: string, psiStrategy?: 'mobile'|'desktop'}} [options]
+   * @param {{psiApiKey?: string, psiApiEndpoint?: string, psiStrategy?: 'mobile'|'desktop', categories?: Array<'performance' | 'accessibility' | 'best-practices' | 'pwa' | 'seo'>}} [options]
```
patrickhulce, line 22: these options are all redundantly prefixed with `psi` because the CLI `collect` command can also use them, I think it might be worth doing at this level for `psiCategories` as well :) (but in `psi-client`, definitely leave without prefix)

WDYT?

patrickhulce (commented):
thanks @jonathantredway this is great!

packages/utils/src/psi-runner.js
jonathantredway: Yeah, sounds very reasonable to stay consistent.

jonathantredway:
@patrickhulce Addressed the feedback ✅, but used the github client to auto-commit your suggestions so it looks like you'll just need to appease googlebot with a consent comment.

patrickhulce:
@googlebot I consent

patrickhulce (approved):
excellent work, thanks @jonathantredway !

PR #726, opened 2021-11-16
feat: upgrade to lighthouse 9.3.0

connorjclark:
include in commit message:

```
BREAKING CHANGE: upgrade to lighthouse v9.3.0
```


connorjclark:
WIP, tests aren't running / snapshots aren't updating as expected locally for me.

nicksrandall:
FYI, lighthouse 9.1 was released so it would be nice if we could get this PR all the way to the latest version

Divlo:
There is now even lighthouse v9.2.0, is there is anything blocking this PR?
Is the failure of the tests are not fixable in CI?

connorjclark:
seems the `node-gyp` errors originate from a parcel v1 dep. https://github.com/parcel-bundler/parcel/issues/5294#issuecomment-720138886

chrillep:
> seems the `node-gyp` errors originate from a parcel v1 dep. [parcel-bundler/parcel#5294 (comment)](https://github.com/parcel-bundler/parcel/issues/5294#issuecomment-720138886)


https://github.com/parcel-bundler/parcel/issues/5294#issuecomment-722542870

packages/cli/test/cli.test.js
```diff
@@ -409,15 +399,6 @@ describe('Lighthouse CI CLI', () => {
                    found: [31mXXXX[0m
               [2mall values: XXXX, XXXX[0m
 
-
-          [31mX[0m  [1mperformance-budget[0m.script.size failure for [1mmaxNumericValue[0m assertion
```
connorjclark, line 413: ref https://github.com/GoogleChrome/lighthouse/issues/13589

packages/utils/src/saved-reports.js
```diff
@@ -51,8 +51,15 @@ function saveLHR(lhr, baseDir = LHCI_DIR) {
...
+  let ReportGenerator;
+  try {
+    // @ts-ignore - lighthouse doesn't publish .d.ts files yet
+    ReportGenerator = require('lighthouse/report/generator/report-generator.js');
+  } catch (_) {}
+  if (!ReportGenerator) {
+    // @ts-ignore - lighthouse doesn't publish .d.ts files yet
+    ReportGenerator = require('lighthouse/lighthouse-core/report/report-generator.js');
+  }
   return ReportGenerator.generateReportHtml(lhr);
```
connorjclark, line 63: TODO: just support latest lighthouse

packages/server/src/ui/routes/build-view/audit-detail/audit-detail-pane.stories.jsx
```diff
@@ -53,6 +57,7 @@ const auditPairs62 = createAuditPairs(lhr62A, lhr62B);
 const auditPairs641 = createAuditPairs(lhr641A, lhr641B);
 const auditPairs700 = createAuditPairs(lhr700A, lhr700B);
 const auditPairs800 = createAuditPairs(lhr800A, lhr800B);
+const auditPairs930 = createAuditPairs(lhr930A, lhr930B);
```
connorjclark, line 60: image

https://github.com/GoogleChrome/lighthouse-ci/blob/main/packages/server/src/ui/routes/build-view/audit-detail/audit-detail-pane.stories.jsx#L166

Not sure why the audit pairs for the other versions don't have this problem, but for the 9.3.0 LHR I'm getting a diff at `screenshot-thumbnails`, as you'd expect. But the viz is non existent so we just print the JSON.

Not sure what the best approach to deal with this is. Is it problematic that this audit is presented this way? How would we prevent it from ever being displayed? Should I just skip it in the `createAuditPairs` like `uses-long-cache-ttl` is skipped?

connorjclark: confirmed this is existing behavior in https://googlechrome.github.io/lighthouse-ci/viewer/

PR #738, opened 2022-01-14
refactor: use esbuild instead of parcel, upgrade to node 14

connorjclark:
Lighthouse 9.x needs Node 14. However, Parcel v1 does not work in 14+, and Parcel v2 proved too difficult a migration[1], so let's use esbuild.

There's a esmodule-only plugin used in this PR that requires an upgrade to Node 14 too, so doing that all in the same PR.

[1] some chat logs
> - means LHCI needs to bump from node 12->14
> - parcel v1 doesnt work in node 14+ (no precompiled gyp binaries + local compile fails)
> - parcel v2 it is!
> - yarn workspaces hoists common deps to root node_modules, which flabbergasts parcel v2
> - ok, fine, yarn workspaces has a "nohoist" thing, let's use that
> - "nohoist" doesn't seem to work at all ... was fixed in yarn 2...


connorjclark:
Also gonna need to upgrade to TS 4 first, but I think we can do that in a separate PR.

PR #739, opened 2022-01-14
chore(deps): upgrade to typescript 4

connorjclark:


packages/cli/src/collect/node-runner.js
```diff
@@ -138,7 +138,7 @@ class LighthouseRunner {
     while (attempts.length < 3) {
       try {
         return await this.run(url, options);
-      } catch (err) {
+      } catch (/** @type {any} */ err) {
```
brendankenny, line 141: if these are always going to be cast to `any`, there's not really any point to the check and it's probably just better to disable it with `"useUnknownInCatchVariables": false` to avoid the noise and potential future issues with type casts. That's what we [do in Lighthouse](https://github.com/GoogleChrome/lighthouse/blob/82772f93645989c9c250c78a0ad10d55f3372beb/tsconfig-base.json#L18).

packages/viewer/src/ui/components/lhci-components.jsx
```diff
@@ -9,8 +9,10 @@ export {LoadingSpinner} from '../../../../server/src/ui/components/loading-spinn
 export {Paper} from '../../../../server/src/ui/components/paper.jsx';
 export {LhrViewerButton} from '../../../../server/src/ui/components/lhr-viewer-button.jsx';
 
+/** @type {string} */
 // @ts-ignore - tsc doesn't get parcel :)
```
brendankenny, line 13:
```suggestion
// @ts-expect-error - tsc doesn't get parcel :)
```
while we're doing this, might as well :)

packages/viewer/src/ui/components/lhci-components.jsx
```diff
@@ -9,8 +9,10 @@ export {LoadingSpinner} from '../../../../server/src/ui/components/loading-spinn
...
 export const CONFETTI_PATH = require('../../../../server/src/ui/routes/project-list/confetti.svg');
 
+/** @type {string} */
 // @ts-ignore - tsc doesn't get parcel :)
```
brendankenny, line 17:
```suggestion
// @ts-expect-error - tsc doesn't get parcel :)
```

brendankenny (approved):
LGTM!

connorjclark:
Flaky ci :(

PR #727, opened 2021-11-16
feat(cli): API requests will respect HTTP_PROXY env variable

nicksrandall:
fixes #574

PR #749, opened 2022-01-28
replace isomorphic-fetch with fetch-with-proxy

abhisheksoni27:
This PR will allow users to use PSI strategy even when they are behind a proxy. They will just have to set the appropriate env variables: `HTTP_PROXY` and `HTTPS_PROXY`


connorjclark:
thanks but going with #727

PR #751, opened 2022-01-31
chore(deps): update lint deps and use implicit configuration

connorjclark:
This fixes tons of editor bugs I had regarding jsx, eslint parsing, etc ...

Because the eslint commands were configured in a non-standard way, eslint+VS code had absolutely no f'in idea what was going on :) I fixed that by making a ton of `.eslintrc.js` files in the proper directories.

Oh, I also removed babel.


.eslintrc.ui.js
```diff
@@ -24,10 +24,11 @@ module.exports = {
         varsIgnorePattern: '(^_$|^h$)',
       },
     ],
+    strict: 'off',
   },
-  parser: 'babel-eslint',
```
paulirish, line 28: this and the 2020 are in the root config.

can you just implicitly adopt those... and i guess flip off `sourceType` and `globalReturn` explicitly in this ui file?

paulirish (approved):
lg!

one nit

.eslintrc.ui.js
connorjclark: removed `parser`.

I can't figure our how `parserOptions` would be merged, so I'll just leave that as-is.

PR #765, opened 2022-02-21
Update package.json to match current server

ptmkenny:
Current package.json after running `npm install --save @lhci/server@latest` according to https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/recipes/heroku-server/README.md


connorjclark:
thanks but went with #782

PR #772, opened 2022-03-18
update dockerfiles to use node 14

connorjclark:
Also had to download some build tools to get the server dockerfile to build. It was failing to find a prebuilt sqlite binary for the node+sqlite version, so has to build from source.


connorjclark:
note to future person (probably myself): google-chrome(-stable) is not published as arm64 yet so you should do this on an intel mac, not m1.

you could build with `--platform=linux/arm64` but can't run to verify it works

Issue #689, opened 2021-09-13
CLS diff not showing on compare screen


Kudja:
**Describe the bug**
I have 2 builds with different CLS
Build 1 - CLS = 0.001
Build 2 - CLS = 0.026
When I compare this 2 builds it will not show CLS in metrics (this works with any threshold)

**To Reproduce**
1. Prepare 2 builds with different CLS
2. Compare them

**Expected behavior**
See CLS diff same way as any other metrics.

**Logs/Screenshots**
I think it's clear w/o any screenshots

**Environment (please complete the following information):**
 - Server - latest docker server
 - Client - latest docker client

patrickhulce:
Thanks for filing @Kudja! I'm not able to reproduce this with just the steps provided though.

![image](https://user-images.githubusercontent.com/2301202/133101418-788a8a00-972e-4c98-acfc-f024dd4d02e8.png)

(Note that the diff is small relative to CLS thresholds so it is represented by "0.0" but it does show up)

Anything additional about your environment or those builds you can share?

Kudja:
Hi @patrickhulce, thanks for your response.

I'm using the official docker-client to gather stats, and docker server to show them. I just installed them a few days ago, so I believe they have the latest versions.
Not sure what else can affect these values.
In config, I'm using perf preset for lighthouse and I have such result in compare
![image](https://user-images.githubusercontent.com/8101375/133144199-50134a4b-c80f-4df2-9bf9-9b574144c74e.png)
CLS is now in Unchanged Audits.
And if I'll open reports for them, then I can see the difference in CLS

If this can help somehow - here are compare link - https://lhci.d1.mageassist.com/app/projects/prilla.kudja.d1.mageassist.com/compare/9a0cac39e56d?baseBuild=14caf986-329f-4b30-af6e-b41b7288a5e6&baseUrl=https%3A%2F%2Fprilla.kudja.d1.mageassist.com%2Fus%2Fnicotine-pouches&compareUrl=https%3A%2F%2Fprilla.kudja.d1.mageassist.com%2Fus%2Fnicotine-pouches

Kudja:
It looks like some rounding issue. When you have very small value of some metric on any side (Base / Compare) it looks like it rounding it to zero and in that case it will always show no diff.

I tested few different builds and If I have CLS = 0 - it's no mateter which value we have on the compare page - it will always show no diff.
If I have CLS = 0.001 then in some cases it will show the diff, in some not
[...]

PR #784, opened 2022-04-18
CLS diff not showing on compare screen

vikasrohit:
#689 CLS diff not showing on compare screen

~~Known issues in the PR:~~

-  ~~I have added duplicate declaration for `SCORE_LEVEL_METRIC_THRESHOLDS` in `score-icon.jsx` because I was not sure what is the ideal way for having it at common place as I didn't see any constant or utils file for the server module. Please let me know, I can fix that quickly.~~
- ~~Not able to resolve the linting error for `const thresholds = SCORE_LEVEL_METRIC_THRESHOLDS[numericDiff.auditId];` in `score-icon.jsx#30`~~

### DIFF
--- a/packages/server/src/ui/components/score-icon.jsx
+++ b/packages/server/src/ui/components/score-icon.jsx
@@ -6,40 +6,11 @@
 
 import {h} from 'preact';
 
-
-/** @type {Record<LHCI.AssertCommand.Budget.TimingMetric, [number, number]>} */
-const SCORE_LEVEL_METRIC_THRESHOLDS = {
-  'first-contentful-paint': [2000, 4000],
-  'first-meaningful-paint': [2000, 4000],
-  'largest-contentful-paint': [2000, 4000],
-  'first-cpu-idle': [3000, 7500],
-  interactive: [3000, 7500],
-  'speed-index': [3000, 6000],
-  'total-blocking-time': [300, 600],
-  'cumulative-layout-shift': [0.1, 0.25],
-  'max-potential-fid': [100, 250],
-};
-
-/** @param {{score: number, diffs?: LHCI.AuditDiff[]}} props */
+/** @param {{score: number}} props */
 export const ScoreIcon = props => {
-  const { score, diffs } = props;
-  let result = score;
-  const numericDiff = diffs && diffs.find(diff => diff.type === 'numericValue');
-  if (numericDiff && numericDiff.type === 'numericValue') {
-    // @ts-ignore
-    const thresholds = SCORE_LEVEL_METRIC_THRESHOLDS[numericDiff.auditId];
-    if (thresholds) {
-      if (numericDiff.compareValue <= thresholds[0]) {
-        result = 0.9;
-      } else if (numericDiff.compareValue <= thresholds[1]) {
-        result = 0.5;
-      } else {
-        result = 0.49;
-      }
-    }
-  }
-  if (result >= 0.9) return <i className="lh-score-pass" />;
-  if (result >= 0.5) return <i className="lh-score-average" />;
+  const score = props.score;
+  if (score >= 0.9) return <i className="lh-score-pass" />;
+  if (score >= 0.5) return <i className="lh-score-average" />;
   return (
     <i className="lh-score-fail">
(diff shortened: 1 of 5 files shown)


packages/utils/src/audit-diff-finder.js
```diff
@@ -668,7 +669,9 @@ function findAuditDiffs(baseAudit, compareAudit, options = {}) {
   // If the only diff found was a numericValue/displayValue diff *AND* it seems like the result was flaky, skip it.
   // The result is likely flaky if the audit passed *OR* it was supposed to have details but no details items changed.
   const isAllPassing = compareAudit.score === 1 && baseAudit.score === 1;
+  const group = getGroupForAuditId(auditId);
   if (
+    group !== 'metrics' && // if metrics group audit is found, don't skip it
```
patrickhulce, line 674: this changes the behavior that was intentional if both were passing, is there a specific case you have in mind for changing this?

packages/server/src/ui/components/score-icon.jsx
```diff
@@ -6,11 +6,40 @@
...
-  if (score >= 0.9) return <i className="lh-score-pass" />;
-  if (score >= 0.5) return <i className="lh-score-average" />;
+  const { score, diffs } = props;
+  let result = score;
+  const numericDiff = diffs && diffs.find(diff => diff.type === 'numericValue');
+  if (numericDiff && numericDiff.type === 'numericValue') {
+    // @ts-ignore
+    const thresholds = SCORE_LEVEL_METRIC_THRESHOLDS[numericDiff.auditId];
+    if (thresholds) {
+      if (numericDiff.compareValue <= thresholds[0]) {
```
patrickhulce, line 32: why are you doing this? the score already has its own thresholds and we want them to match the actual score.

patrickhulce (commented):
thanks @vikasrohit!  a few questions :)

packages/utils/src/audit-diff-finder.js
vikasrohit: I think @patrickhulce we agreed [here](https://github.com/GoogleChrome/lighthouse-ci/issues/689#issuecomment-948556023) on skipping this check when the group id is `metrics`.

packages/server/src/ui/components/score-icon.jsx
vikasrohit: With current implementation, CLS audit always get score of 1 for me and because it is always 1, it always show passing icon for CLS audit diff. I may be wrong with my observation for having score 1 for CLS. Please correct me if my observation is wrong here.

vikasrohit: @patrickhulce I just tried again with a page which has bad CLS score and I can see non 1 score for CLS, which means my assumption was wrong there. Let me fix the changes accordingly and come back.

vikasrohit: @patrickhulce I have reverted the changes for this file as I have verified that scores are working fine and we don't need to handle the thresholds separately for CLS.

packages/utils/src/audit-diff-finder.js
patrickhulce: Ah thanks for the reminder it had been awhile :) sg!

packages/server/src/ui/routes/build-view/audit-list/audit-group.jsx
```diff
@@ -58,7 +58,10 @@ const ChangedAuditList = props => {
             onClick={() => props.setSelectedAuditId(audit.id || null)}
           >
             <div className="audit-group__audit-score">
-              <ScoreIcon score={audit.score || 0} />
+              <ScoreIcon
```
patrickhulce, line 61: I think we can revert these too now :)

patrickhulce (commented):
thanks for the contribution @vikasrohit! just one more revert and I think we're good here

packages/server/src/ui/routes/build-view/audit-list/audit-group.jsx
vikasrohit: Done. Thanks for noticing it, my bad I missed it with last revert.

patrickhulce (approved):

PR #785, opened 2022-04-29
feat: extract type defs to assert.d.ts

mxdvl:
## What does this change? 

Moves the type definitions of assertions into the main declaration file, under the `LHCI.AssertResults` namespace.

## Why?

This allows the output JSON file to have an expected type. This is especially useful in Deno, where you can do:

```ts
// With a version number that’s not actually v0.9.0
import "https://raw.githubusercontent.com/GoogleChrome/lighthouse-ci/v0.9.0/types/assert.d.ts";

const results: LHCI.AssertResults.AssertionResult[] = JSON.parse(
	Deno.readTextFileSync(`.lighthouseci/assertion-results.json`)
);

```


connorjclark:
Seems cool to me. Thanks.

PR #824, opened 2022-08-21
Add wait to open

mauron85:
prevent closing shell before report is opened in the browser


connorjclark:
Seems fine, but can you help me understand why this matters to you if the command doesn't wait?

PR #825, opened 2022-08-21
Update configuration.md

mauron85:
Update lhci open instruction to mention that URL must be provided to open to work.


connorjclark:
I believe a parameter here is optional, and otherwise it will open all available (median) reports.

PR #835, opened 2022-10-22
feat(cli): Added url hash to differentiate between urls

jaspritt-hora:
### ADDED:
- `hash` parameter, while generating reports for apps using #/ 
- Test case added for the same

### NOTE: 
- This change is in reference to [this](https://github.com/GoogleChrome/lighthouse-ci/discussions/834) thread, as per discussion with @patrickhulce.

### DIFF
--- a/packages/cli/src/upload/upload.js
+++ b/packages/cli/src/upload/upload.js
@@ -543,6 +543,7 @@ async function runFilesystemTarget(options) {
     const context = {
       hostname: url.hostname,
       pathname: url.pathname,
+      hash: url.hash,
       date: fetchTimeDate.toISOString().replace(/T.*/, ''),
       datetime: fetchTimeDate
         .toISOString()
--- a/packages/cli/test/upload-url-hash.test.js
+++ b/packages/cli/test/upload-url-hash.test.js
@@ -1,0 +1,116 @@
+/**
+ * @license Copyright 2019 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+/* eslint-env jest */
+
+const fs = require('fs');
+const path = require('path');
+const rimraf = require('rimraf');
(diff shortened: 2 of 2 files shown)


jaspritt-hora:
@patrickhulce - Changes are done. Let me know, please review and let me know, if anything is required.

packages/cli/src/upload/upload.js
```diff
@@ -542,7 +542,7 @@ async function runFilesystemTarget(options) {
     const fetchTimeDate = new Date(new Date(lhr.fetchTime).getTime() || Date.now());
     const context = {
       hostname: url.hostname,
-      pathname: url.pathname,
+      pathname: url.hash + url.pathname,
```
patrickhulce, line 545: we don't want to merge these two, we would want them to be separate
```suggestion
      pathname: url.pathname,
      hash: url.hash,
```

patrickhulce (changes requested):
thanks @jaspritt-hora!

could you add a test case here?

packages/cli/src/upload/upload.js
jaspritt-hora: @patrickhulce - Regarding the test case, we want to add the test case for both `collect` & `upload` or just the latter?

patrickhulce: Collect isn't being touched here, just upload

jaspritt-hora: @patrickhulce - Sounds good! I have made the suggested change and written the test case in a new file and it is working as expected. 

However, I have a question, NOT all the test cases are running on my machine. 
Refer following screenshot - 

The unit test case file, I have written is - `upload-url-hash.test.js`. The test for this has passed. 
image

However, `NOT ALL` the test suites are passing. Refer following screenshots - 
image

Can I go ahead and commit the changes ? 
OR 
Do I need the entire test suite to pass?

patrickhulce: don't worry about the whole test suite for now, some are flaky and/or require other setup on your machine like mysql, if upload tests are passing you should be a-ok

jaspritt-hora: Perfect!

jaspritt-hora: @patrickhulce - Can you the checks again. The issue was related to syncing of the directories, I have added `rimraf` and the required code for the same. The test case is running successfully now, had 3 consecutive successful runs.

jaspritt-hora: @patrickhulce - Checked the raw logs for both `windows` and `macOS` checks that have failed, in that - the test case written for the PR has passed successfully.

The test case file name is - `upload-url-hash.test.js`.
Let me know, if there anything actionable at my end.

patrickhulce (approved):

PR #849, opened 2022-12-05
feat(cli): use requestedUrl instead of finalUrl for upload context naming

jaspritt-hora:
**CHANGED**:

- Using `requestedUrl` instead of `finalUrl` for `runFilesystemTarget ` function.
- Updated the test cases for `target=filesystem` in `upload.test.js`
- Updated the test cases for `upload-url-hash.test.js`

**SUPPORTING INFO**:
- This in relation to the ongoing discussion [here](https://github.com/GoogleChrome/lighthouse-ci/discussions/834#discussioncomment-4256769)
- #835 is part I of the implementation


jaspritt-hora:
@connorjclark / @patrickhulce  - Please review the changes. I have made changes wrt `target=filesystem` only for now.

patrickhulce (commented):
technically breaking but seems fine to me 🤷

packages/cli/test/upload-url-hash.test.js
```diff
@@ -1,5 +1,5 @@
 /**
- * @license Copyright 2019 Google Inc. All Rights Reserved.
```
connorjclark, line 2: please revert - as a policy we don't update these years

jaspritt-hora: @connorjclark - Done and I have squashed the commits as well

connorjclark:
Thanks @jaspritt-hora!

PR #871, opened 2023-02-15
chore(dep): upgrade to puppeteer 19.7.1

connorjclark:
- upgrade to puppeteer 19.7.1
- upgrade to jest 29.4.3, to get support for "exports" field needed for puppeteer to work
- port `52427` is used by a system service on Mac... so can't use that in tests

There's the usual text rendering deltas visible in the image snapshots, but there are some things that could be considered bugs to fix:

- `project-dashboard-mixed-v-5-v-6-test-js-project-dashboard-render-the-dashboard-should-look-correct-1-snap.png`: oddly has a slightly different y-values on the last two points of the first graph. In fact, _every_ graph seems to shift inexplicable on the last bit. I bisected this to puppeteer 19.0.0, which is Chromium range: 109.0.5412.0...110.0.5479.0
- Dates are now including seconds
- `project-dashboard-mixed-v-5-v-6-test-js-project-dashboard-render-the-dashboard-should-look-correct-1-snap.png`: why the extra space in the date string?
- `simple-comparison-test-js-viewer-simple-comparison-render-the-landing-route-should-look-correct-1-snap.png`: completely lost the "Base" label

Fingers crossed this makes tests less flaky.


connorjclark:
I suspect there was no change to the actual graph rendering, at least no within the range we have here... so few pixels in a line that it wouldn't have triggered the snapshot failure threshold, but the fact that the timestamp regressed (due to toLocaleString in Chrome now using a NBSP instead of a space...), was finally enough to fail the test.

Perhaps the underlying data changed since the snapshot was last updated.

PR #877, opened 2023-02-25
feat(server): allow suppressing bodyParser to use own middleware

jorisdugue:
Hi thanks for this project 😄 

This PR allows to use **fastify** in custom middleware. Without this PR you cannot wrap this express server due to **bodyParser**

### DIFF
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -29,7 +29,7 @@ const DIST_FOLDER = path.join(__dirname, '../dist');
  * @return {Promise<{app: Parameters<typeof createHttpServer>[1], storageMethod: StorageMethod}>}
  */
 async function createApp(options) {
-  const {storage} = options;
+  const {storage, useBodyParser} = options;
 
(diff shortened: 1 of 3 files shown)


connorjclark (approved):
Seems reasonable to me. Thanks, especially for writing an example!

packages/server/src/server.js
```diff
@@ -43,9 +43,12 @@ async function createApp(options) {
   // While LHCI should be served behind nginx/apache that handles compression, it won't always be.
   app.use(compression());
 
-  // 1. Support large payloads because LHRs are big.
-  // 2. Support JSON primitives because `PUT /builds/<id>/lifecycle "sealed"`
-  app.use(bodyParser.json({limit: '10mb', strict: false}));
+  if (typeof useBodyParser === 'undefined' || useBodyParser) {
```
paulirish, line 46: little late but.. maybe we flip the flag so we don't have to do this kinda check?

`dontUseBodyParser` or `skipBodyParser` ?

PR #881, opened 2023-03-03
fix(server): skip needless/verbose audit diffs

paulirish:
A number of audits currently get full diff rendering but its nonsensical to diff them. (main-thread-tasks is an obvious one).
This introduces a blocklist for those audits. 

It also dramatically improves the performance when viewing two lhrs that have lots of tasks.


paulirish:
@connorjclark this one work for you too?

PR #883, opened 2023-03-07
audit diff: use good keys for nodes and taptargets

paulirish:
minor stuff.


adamraine (approved):

paulirish:
i reviewed the latest image snapshots. they seem good.

PR #895, opened 2023-04-08
detect 11ty's output dir as `staticDistDir`

koddsson:
Hey 👋🏻 

I was surprised when Lighthouse CI didn't detect my built page when setting up a new build. The error message was really helpful, so I got up and running quickly, but I figured I'd ask if you'd consider adding `_site` to the list of automatically detect `staticDistDir` directories.

I figured that it would be a low-risk change, since I suspect that there aren't many directories named `_site` in other systems.

Let me know what you think :)

## References
https://www.11ty.dev/docs/config/#output-directory


paulirish (approved):
Good idea. Yeah sounds great.

Thanks for the A+ PR

PR #898, opened 2023-04-12
fix: update render snapshots

adamraine:


paulirish (approved):
We see a few datapoint positions change in packages/server/test/e2e/__image_snapshots__/project-dashboard-mixed-v-5-v-6-test-js-project-dashboard-render-the-dashboard-should-look-correct-1-snap.png .. and that's somewhat unexplainable. but we're moving ahead anyway

PR #907, opened 2023-04-27
collect: use new headless mode

brendankenny:
Switch to --headless=new when `headful` isn't requested.

This is better in general, but should also fix bfcache results which have issues with the old headless mode (see https://github.com/treosh/lighthouse-ci-action/issues/126, https://github.com/WPO-Foundation/wptagent/pull/605).

I'd like to add a test that the bfcache audit isn't blocked by getting the "Back/forward cache is not supported by delegate" failure anymore, but I haven't found a good place to add that. Happy to take any suggestions!

### DIFF
--- a/packages/cli/src/collect/node-runner.js
+++ b/packages/cli/src/collect/node-runner.js
@@ -45,7 +45,7 @@ class LighthouseRunner {
     const settings = options.settings || {};
     const chromeFlags = options.settings && options.settings.chromeFlags;
     let chromeFlagsAsString = chromeFlags || '';
-    if (!options.headful) chromeFlagsAsString += ' --headless';
+    if (!options.headful) chromeFlagsAsString += ' --headless=new';
     if (chromeFlagsAsString) settings.chromeFlags = chromeFlagsAsString;
 
     // Make sure we're not passing something that will ruin our runner.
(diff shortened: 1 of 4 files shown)


packages/cli/test/collect-puppeteer.test.js
```diff
@@ -78,7 +78,7 @@ describe('Lighthouse CI collect CLI with puppeteer', () => {
     const files = fs.readdirSync(path.join(autorunDir, '.lighthouseci'));
     const report = files.find(file => /lhr.*\.json$/.test(file));
     const lhr = JSON.parse(fs.readFileSync(path.join(autorunDir, '.lighthouseci', report)));
-    expect(lhr.userAgent).toContain('HeadlessChrome/111.0.5555.0'); // make sure the right chrome was used
+    expect(lhr.userAgent).toContain('Chrome/111.0.0.0'); // make sure the right chrome was used
```
brendankenny, line 81: Needed because `--headless=new` no longer announces itself in the UA string.

From #336 the condition being asserted seems to be that puppeteer's Chrome was used, not that it was headless in particular (that's just the default without `headful: true` in the options), so this change seems ok.

adamraine (approved):

gmerwan (approved):

rogerfernandes:
Guys, any chance this Pull Request will be merged soon?

PR #927, opened 2023-08-05
Instruct users to install Lighthouse CI 0.12

koddsson:
Yo! Figured this should be updated since we're on version 0.12 now.


adamraine (approved):

adamraine:
Thanks, good catch!

PR #985, opened 2023-12-14
chore: set jest max workers to 1

connorjclark:
There are a few tests that I suspect squash each other when run in parallel. The previous value here set max to 2 (probably because the longest test by far - storybook - is not impacted), but that could still result in a problem.

ref #981


paulirish (approved):
sgtm

PR #991, opened 2023-12-14
feat: upgrade to lighthouse 11.4.0

connorjclark:


adamraine (commented):
Should we update these tests as well?

https://github.com/GoogleChrome/lighthouse-ci/blob/main/packages/server/test/api/statistic-definitions.test.js
https://github.com/GoogleChrome/lighthouse-ci/blob/main/packages/server/src/ui/routes/build-view/lhr-comparison.stories.jsx

PR #997, opened 2023-12-22
[Feature] set up additional config `totalBuildsToKeep` for deleteOldBuilds Job

metalix2:
## Background

We've found over time that different Projects have more activity than others. We've seen the MAX history for a specific runURL can be as little as 3 weeks ago. Where less frequently tested projects can have a history of 3 Months within the same amount of builds shown in the graph. Due to this wide range we have 2 months worth of builds for the busier project that are unviewable in the UI. 

To keep storage optimisied we think the option to be able to keep a MAX amount of builds (like 150 or whatever is available to be viewed in the UI) would be more suitable for the different project types.  We don't want to limit all projects to three weeks as they may not make as many frequent changes.

## Changes

- Added server config `totalBuildsToKeep` as part of `deleteOldBuildsCron`
- Added `findRemainingBuilds` method to `storage-method` to return builds found after `totalBuildsToKeep`
- Changed Configuration Docs for `totalBuildsToKeep`
- Changed Tests and Logic around deleteOldBuilds
- Changed `maxAgeInDays` to be an optional config option. Now either `totalBuildsToKeep` or `maxAgeInDays` can be used or both?

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1050,6 +1050,10 @@ The [cron-style](https://crontab.guru/) schedule on which to collect results.
 
 A number of days to keep a records.
 
+##### `deleteOldBuildsCron[i].totalBuildsToKeep`
+
+A number of total builds to be kept, any builds found there after given will be deleted.
+
 ##### `deleteOldBuildsCron[i].onlyBranches`
 
 Array of branch names, which should be deleted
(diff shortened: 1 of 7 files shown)


metalix2:
Another alternative is to set target projects in the cronJob that way we could create jobs per project with reduced `maxAgeInDays`.  Let me know your preference and I'll update 👍

metalix2:
Any thoughts on this?

PR #1004, opened 2024-01-15
Changes to use CLI args to configure VIEWER_ORIGIN for lighthouse server

bkjam:
Hi,

I like the solution in #637 which allows us to change the default self-hosted lighthouse report viewer. However, I noticed that the `$VIEWER_ORIGIN` environment variable works only if you build a custom `@lhci/server` package. Hence, I made some modifications to configure `$VIEWER_ORIGIN` for `@lhci/server` via the CLI args.

I changed the following things:

- added an optional `--viewer.origin` cli args for server (Default: https://googlechrome.github.io) 
- added `/v1/viewer/origin` endpoint that retrieves the `viewer.origin` value
- updated the server docs

This may not be the most optimal solution, so I would appreciate any feedback on my approach and code changes. Thank you :)

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -171,3 +171,19 @@ const lhci = require('@lhci/server');
 ### Firewall Rules
 
 You can also protect your server through firewall rules to prevent it from being accessed from outside your internal network. Refer to your infrastructure provider's documentation on how to setup firewall rules to block external IP addresses from accessing the server. Don't forget to allow your CI machines!
+
+### Self-Hosted Lighthouse Report Viewer
+
+If you are hosting your own lighthouse report viewer instead of using the [default viewer](https://googlechrome.github.io/lighthouse/viewer), you can add `--viewer.origin` to your lighthouse server configuration. Eg.
+
+```js
+const {createServer} = require('@lhci/server');
+
+console.log('Starting server...');
+createServer({
+  ...
+  viewer: {
+    origin: 'https://viewer-url' // Default: https://googlechrome.github.io
+  }
+}).then(({port}) => console.log('LHCI listening on port', port));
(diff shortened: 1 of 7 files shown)


brendangadd:
> Do you mean that it only works if you host your own LHCI server? As I understand, you don't need to build anything yourself - the package from npm for @lhci/server will work if you run it with VIEWER_ORIGIN env defined.

@connorjclark: We've been working on using our own report viewer deployment as well, so maybe I can shed some light.

The challenge is that `process.env.VIEWER_ORIGIN` in `LhrViewerLink`, being client-side, is evaluated at build-time and not at runtime. Looking at the build result from the `@lhci/server` package, we can see that the default value is unconditional.

```sh
$ cat node_modules/@lhci/server/dist/chunks/entry-*.js | grep -oP 'function[^}]*googlechrome.github.io.{40}'
function La(e){let t="https://googlechrome.github.io";window.addEventListener("message",func
```

This means you have to build `@lhci/server` yourself if you want to view the LH reports in a non-default location (e.g. a self-hosted viewer). This PR moves resolution of the viewer URL server-side so that it can be set at runtime and the client can look it up via the API.

By the looks of it, @bkjam preserved priority of the build-time env variable, so people who _are_ building their own package with a custom viewer origin should get consistent behaviour. Though I think that means that packages having a build-time origin will silently ignore any server-side configuration.

In any case, this is a feature that my team would use! :)

bkjam:
Thanks so much, @connorjclark and @brendangadd, for helping with this pull request! The rationale is as explained by @brendangadd. Apologies for missing the comments for so long—I’ll do better going forward. Thanks again for your patience!

PR #1024, opened 2024-04-09
feat(cli): add --lhr to assert command to load LHRs from anywhere

connorjclark:
ref https://github.com/GoogleChrome/lighthouse/issues/15203

Enables usage like this:

```
node packages/cli/src/cli.js -no-lighthouserc assert \
  --lhr /Users/cjamcl/src/lighthouse/core/test/results/sample_v2.json \
  --budgetsFile=packages/cli/test/fixtures/budgets.json
```

### DIFF
--- a/packages/cli/src/assert/assert.js
+++ b/packages/cli/src/assert/assert.js
@@ -32,6 +32,10 @@ function buildCommand(yargs) {
       type: 'boolean',
       description: 'Whether to include the results of passed assertions in the output.',
     },
+    lhr: {
+      description:
+        'Path to LHRs (either a folder or a single file path). Not recursive. If not provided, .lighthouseci is used',
+    },
   });
 }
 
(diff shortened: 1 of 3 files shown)


paulirish (approved):
nice

fahmij8:
hey team, sorry to discuss here but is this still working?
I tested it out, but it always says `Checking assertions against 0 URL(s), 0 total run(s)`
image

packages/cli/src/assert/assert.js
```diff
@@ -53,7 +57,7 @@ async function runCommand(options) {
   // If we have a budgets file, convert it to our assertions format.
   if (budgetsFile) options = await convertBudgetsToAssertions(readBudgets(budgetsFile));
 
-  const lhrs = loadSavedLHRs().map(json => JSON.parse(json));
+  const lhrs = loadSavedLHRs(options.lhr).map(json => JSON.parse(json));
```
fahmij8, line 60: it looks like if we pass `options.lhr` it will be gone cause it was replaced by the `convertBudgetsToAssertions` returned value.

if my understanding is correct & this is a vallid issue, I would love to create PR to fix this

fahmij8: Thanks! fixes coming right up

fahmij8: The fixes are at #1062, @connorjclark.
Could you kindly review it at your convenience?

PR #1035, opened 2024-05-07
feat: upgrade to lighthouse 12.0.0

connorjclark:


RaHehl:
@connorjclark  Do you already know when there will be a release for it?

PR #1068, opened 2024-08-03
Don't mutate user settings

### LINKED ISSUE #1022: Documentation: How to test a site in dark mode
**Is your feature request related to a problem? Please describe.**

The accessibility assessment checks contrast ratios.  Websites often provide a light theme and a dark theme, and it is common to for the site to automatically set the theme based on the browser/OS preference.

**Describe the solution you'd like**

It would be great to have an easy way to run "dark mode" tests against sites that let the browser/OS select the theme.

**Describe alternatives you've considered**

I read through the documentation but it seems quite complicated to set up, and searching for "dark mode" yielded no results.

### LINKED ISSUE #1067: Second run and beyond falls back to old Chrome headless mode
**Describe the bug**

The Chrome `--headless=new` option is only effective on the first run because of repeated `--headless=new --headless=new ...` for the other runs, which apparently triggers the old headless mode.

**To Reproduce**
1. Pick a URL that reports a significantly lower or higher performance with the `--headless=new` mode compared to `--headless`.
  * I use https://dev.d33869rf57mkjo.amplifyapp.com/ as an example here (found in https://github.com/GoogleChrome/lighthouse/issues/6708).
  * http://www.lumme.info/ (from https://github.com/GoogleChrome/lighthouse/issues/14957) should also show an effect (only terminate with `--headless=new`)
2. Use this `.lighthouserc.yml`:
```yml
ci:
  assert:
    assertions:
      categories:performance: [error, minScore: 0]
    includePassedAssertions: true
  collect:
    url:
      - https://dev.d33869rf57mkjo.amplifyapp.com/
  upload:
    outputDir: .lighthouseci/results
    target: filesystem
```
3. `npx lhci autorun`
4. Check the three reported values
  * The first value matches the score of the browser devtools Lighthouse.
  * The second and third score are different from the first.

```
Checking assertions against 1 URL(s), 3 total run(s)

1 result(s) for https://dev.d33869rf57mkjo.amplifyapp.com/ :

  ✅  categories.performance passing for minScore assertion
        expected: >=0
           found: 1
      all values: 1, 0.73, 0.72
```

**Expected behavior**

* All three scores should be similar.
* All three scores should match the browser devtools Lighthouse result.

The expected behavior can be enforced by adding under `collect:`:
```yaml
    headful: true
    settings:
      chromeFlags: --headless=new
```

```
Checking assertions against 1 URL(s), 3 total run(s)

1 result(s) for https://dev.d33869rf57mkjo.amplifyapp.com/ :

  ✅  categories.performance passing for minScore assertion
        expected: >=0
           found: 1
      all values: 1, 1, 1
```

 - OS: macOS 14.5
 - Browser: Chrome
 - Version: 127.0.6533.89

**Additional context**

`console.log({chromeFlagsAsString})` in `static computeArgumentsAndCleanup(url, options)` logs the problem.

dcsaszar:
Fixes #1067
Closes #1022


connorjclark:
thanks!

PR #1069, opened 2024-08-27
chore(wizard): change default branch default to main

mmeester:
Since the broad adaption and change of policy to have the main branch to be called main, it's time to set default in the wizard to main as well


packages/cli/src/wizard/wizard.js
MeAgainEricaccea, line 1: master cannot define function or replace original repository creation

MeAgainEricaccea (changes requested):

connorjclark:
thanks!

PR #1083, opened 2024-11-12
Update audit IDs reference link in configuration documentation

thibaudcolas:
Updates the [default Lighthouse config](https://github.com/GoogleChrome/lighthouse/blob/v12.2.1/core/config/default-config.js#L381-L432) link to the latest release. The file has changed locations, and there are a lot of audit changes (going from 33 to 49 just for the performance category), so I thought this would be worth updating.

If this feels useful I’m happy to also update other links to the Lighthouse source elsewhere in the documentation, though none of the ones I stumbled upon showed such significant changes as this one.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -690,7 +690,7 @@ Options:
 
 #### `assertions`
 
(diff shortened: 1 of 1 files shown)


connorjclark:
covered [here](https://github.com/GoogleChrome/lighthouse-ci/pull/1091) (but thanks all the same)

PR #1084, opened 2024-11-13
improve filename sanitization for hash routes

Awsum-Al:
This enhancement modifies the filename sanitization process to handle hash-based routing more elegantly. By removing leading underscores from sanitized values, we create cleaner and more readable filenames for Single Page Applications using hash-based navigation.

Changes:

- Updated getFileOutputPath sanitization to remove leading underscores
- Maintains all existing functionality while improving readability
- Particularly beneficial for React, Vue, and other SPA frameworks
- Preserves path information without unnecessary prefixes

Example transformation: Before: _dashboard-2024_11_13.report.html After: dashboard-2024_11_13.report.html


connorjclark:
thanks!

Issue #1108, opened 2025-06-20
[Bug] 0.15.0 Report Viewer Malformed URL


JoelJamesDuff:
**Describe the bug**
Introduced in v0.15.0 via #1004 
Attempting to open the Lighthouse Report Viewer may result in a malformed URL.

**To Reproduce**
1. Configure lhci server with no VIEWER_ORIGIN environment variable.
2. Click on the lhr-viewer-link Report button. (A breakpoint helps here to freeze the newly opened tab before it redirects.)
3. Observe the incorrect URL (detailed below) 

**Expected behavior**
The Lighthouse Report Viewer opens at `https://googlechrome.github.io/lighthouse/viewer`

**Actual behaviour**
A new tab opens at `{host}/app/projects/{projectSlug}/[object%20Object]/lighthouse/viewer`

**Explanation**
The `/viewer/origin` endpoint and consequently `api.getViewerOrigin()` return the object `{viewerOrigin: "https://googlechrome.github.io"}` instead of just a pure string. The viewer origin string is never extracted from this object and at lhr-viewer-link.jsx line 38, the object is implicitly converted into a string with ``const popup = window.open(`${VIEWER_ORIGIN}/lighthouse/viewer`, windowName);``

The resulting URL handed to window.open is `[object Object]/lighthouse/viewer` which js resolves by appending it to the end of the URL of the current page.

andys8:
Fix is merged. Hope it gets released soon :) 

https://github.com/GoogleChrome/lighthouse-ci/pull/1109

PR #1109, opened 2025-06-24
fix: use viewer origin as string instead of object

akhilbv:
api.getViewerOrigin() returns the object {"viewerOrigin": "<URL>"} instead of the string value. This causes malformed report URL. Clicking on the report link redirects to the home page because of this. The URL constructs as ´{projectURL}/[object%20Object]/lighthouse/viewer´.  This fix replaces the [object%20Object] with actual string value.

### DIFF
--- a/packages/server/src/ui/components/lhr-viewer-link.jsx
+++ b/packages/server/src/ui/components/lhr-viewer-link.jsx
@@ -15,7 +15,7 @@ import './lhr-viewer-link.css';
 /** @param {LH.Result} lhr */
 export async function openLhrInClassicViewer(lhr) {
   // Allows for self-hosted viewer uri instead of linking externally
-  const cliViewerOrigin = await api.getViewerOrigin();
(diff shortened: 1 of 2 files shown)


akhilbv:
Fixes [this issue](https://github.com/GoogleChrome/lighthouse-ci/issues/1108)

paulirish (approved):

paulirish:
legit thanks.

akhilbv:
@paulirish Thanks for merging the changes.

Do you know when this will be released? The report functionality is broken for the docker-image('patrickhulce/lhci-server') users

andys8:
Thanks for the update! I am really looking forward to the fix. It would be fantastic to see a release in the near future.

After the release the [docker image](https://github.com/GoogleChrome/lighthouse-ci/blob/76a49c7cc26cfc6dcff4248e1f170efb845245bb/docs/recipes/docker-server/package.json#L8) needs to be updated.

 @connorjclark

kareem55555 (commented):
help me

PR #1111, opened 2025-08-20
Use latest GitHub Action for checkout

fulldecent:
Just a quick bump. Thank you.

---

Documentation reference for supporting this change:

https://github.com/actions/checkout/releases

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -84,7 +84,7 @@ jobs:
     name: Lighthouse
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Use Node.js 16.x
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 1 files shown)


xavierfoucrier:
Unless I am mistaken, currently it's not possible since **LHCI** doesn't support NodeJS 24 as well.. which is the minimum requirement for using [@actions/checkout@v5](https://github.com/actions/checkout/releases/tag/v5.0.0).

terrabitz:
As of November 2025, it looks like actions/checkout is v6 now: https://github.com/actions/checkout/releases/tag/v6.0.0

Also @xavierfoucrier that shouldn't matter here. The supported version you see on `actions/checkout` refers to the Node.js runtime required to build that action, not the one we install and use as part of the CI job. They should be completely orthogonal.

xavierfoucrier:
@terrabitz I finally dive and fix the issue I was getting with **LHCI**, it's clearly not related to `actions/checkout` 😉 

> I have marked my previous comment as resolved.

PR #1116, opened 2025-10-01
refactor: connect to existing Chrome instance

### LINKED ISSUE #16259: CLI flag to choose whether to open a new lighthouse page or not
**Feature request summary**

A CLI flag (something like `--reuse-page`) to re-use the existing page that was running on the browser that puppeteer connects to.
The page can be selected from `await lhBrowser.pages()`.

Yes

**What is the motivation or use case for changing this?**
It will help run lighthouse programmatically inside applications like CEF and Electron where opening a new page (by default) causes a failure when connecting to it through remote debugging port.

**How is this beneficial to Lighthouse?**
It will widen the audience to which lighthouse can be used and increase the tool's users.

Mentioned the changes in the screenshot:
lh module changes

khushmanvar:
- Connect to existing instance
- Remove launch options
- fixes [issue #16259](https://github.com/GoogleChrome/lighthouse/issues/16259)

### DIFF
--- a/packages/cli/src/collect/puppeteer-manager.js
+++ b/packages/cli/src/collect/puppeteer-manager.js
@@ -58,19 +58,7 @@ class PuppeteerManager {
       throw new Error(`Unable to require 'puppeteer' for script, have you run 'npm i puppeteer'?`);
     }
 
-    /** @type {import('puppeteer').PuppeteerLaunchOptions} */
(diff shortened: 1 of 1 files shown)


connorjclark:
um, the idea was to do something like this behind a flag. This totally destroys all existing usage of this library.

Issue #1131, opened 2026-02-26
Lighthouse is using a vulnerable version of inquirer and tmp


kennethbigler:
**Describe the bug**
Lighthouse is using a vulnerable version of inquirer and tmp.

**To Reproduce**
1. run `npm i`
2. run `npm audit fix`

**Expected behavior**
There should be 0 vulnerabilities when running `npm audit`.

**Logs/Screenshots**
```
# npm audit report

tmp  <=0.2.3
tmp allows arbitrary temporary file / directory write via symbolic link `dir` parameter - https://github.com/advisories/GHSA-52f5-9888-hmc6
fix available via `npm audit fix --force`
Will install @lhci/cli@0.1.0, which is a breaking change
node_modules/external-editor/node_modules/tmp
node_modules/tmp
  @lhci/cli  *
  Depends on vulnerable versions of inquirer
  Depends on vulnerable versions of tmp
  node_modules/@lhci/cli
  external-editor  >=1.1.1
  Depends on vulnerable versions of tmp
  node_modules/external-editor
    inquirer  3.0.0 - 8.2.6 || 9.0.0 - 9.3.7
    Depends on vulnerable versions of external-editor
    node_modules/gts/node_modules/inquirer
    node_modules/inquirer
      gts  *
      Depends on vulnerable versions of inquirer
      node_modules/gts

5 low severity vulnerabilities
```

**Environment (please complete the following information):**
 - OS: Mac Tahoe 26.3
 - Browser: N/A
 - Lighthouse Version: v0.15.1
 - Node Version: v24.11.0
 - npm Version: v11.9.0

**Additional context**
N/A

Issue #393, opened 2020-07-30
npm audit issues


vingogo:
![image](https://user-images.githubusercontent.com/27344004/88947952-f4f8d980-d299-11ea-87f4-bca937dcdfe0.png)
![image](https://user-images.githubusercontent.com/27344004/88948038-14900200-d29a-11ea-8b53-efb9c24053f4.png)
![image](https://user-images.githubusercontent.com/27344004/88948079-2376b480-d29a-11ea-8495-baf6fca261a9.png)

patrickhulce:
Thanks for filing @vingogo ! 

In principle we won't really take action to address transitive `low` severity warnings like prototype pollution for a CLI. 

The `high` severity issues are blocked by https://github.com/GoogleChrome/lighthouse/issues/10404.

PR #1139, opened 2026-04-29
fix: replace uuid dependency with crypto.randomBytes

piotrekwitkowski:
## Summary
- Removes the `uuid` package from `@lhci/cli` and `@lhci/server` to eliminate [GHSA-w5hq-g745-h8pq](https://github.com/advisories/GHSA-w5hq-g745-h8pq) (moderate severity, uuid <14.0.0)
- Adds a `generateUUID()` utility in `@lhci/utils/src/uuid.js` using `crypto.randomBytes` with manual UUID v4 bit formatting
- Compatible with Node 18 (no dependency on `crypto.randomUUID` which requires Node 19+)

## Related
- #1131
- #393

## Test plan
- `yarn jest packages/server/test/api/storage/storage-method.test.js` passes
- `yarn jest packages/utils/test` passes (pre-existing failures in `build-context.test.js` are unrelated)

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -25,7 +25,6 @@
     "lighthouse-logger": "1.2.0",
     "open": "^7.1.0",
     "tmp": "^0.1.0",
-    "uuid": "^8.3.1",
     "yargs": "^15.4.1",
     "yargs-parser": "^13.1.2"
   }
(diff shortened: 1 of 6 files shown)


deborahgu:
Hi, I'm wondering if this is likely to get merged? I was going to submit a PR to bump `uuid`, but if it's just going to be removed I won't bother.

